PR #24, opened 2015-07-16
Add edit mode and feature drag

kelvinabrokwa:
drag

Implements an extensible edit mode for features and uses it for feature dragging, re: https://github.com/mapbox/gl-draw/issues/23, https://github.com/mapbox/gl-draw/issues/27, part of https://github.com/mapbox/gl-draw/issues/26


tristen:
Awesome! Just tested and I'm seeing a couple things:

- Dragging should be disabled when you are drawing. i.e. If the marker button is toggled the state of the editor is in draw mode, not edit.

- A feature should be clicked first (when out of draw mode) to put itself into edit mode. Then mousedown and drag is enabled on the selected edit mode feature.

kelvinabrokwa:
square_drag

kelvinabrokwa:
line_drag

kelvinabrokwa:
@tristen @tmcw please review this when you can. The basic idea here is to setup a way to edit features (and the structure is used to implement translation (drag) in this PR as well).

The highlights:
- [`onClick` behavior][1] for toggling edit mode and switching between features to edit
- [entering and exiting][2] edit mode
- a simple [edit store][3] for keeping track of in progress edits
- [translation code][4] for drag feature ([points][5] implement a simpler one)
- [edit style][6] for polygon

[1]: https://github.com/mapbox/gl-draw/blob/drag/src/draw.js#L143-L169
[2]: https://github.com/mapbox/gl-draw/blob/drag/src/draw.js#L171-L188
[3]: https://github.com/mapbox/gl-draw/blob/drag/src/edit_store.js
[4]: https://github.com/mapbox/gl-draw/blob/drag/src/handlers/handlers.js#L76-L115
[5]: https://github.com/mapbox/gl-draw/blob/drag/src/handlers/point.js#L29-L48
[6]: https://github.com/mapbox/gl-draw/blob/drag/src/theme/edit.js#L31-L56

src/draw.js
```diff
@@ -91,47 +108,136 @@ Draw.prototype = extend(Control, {
...
+
+      var feature = features[0];
+      if (this.editId && this.editId !== feature.properties._drawid) this._exitEdit();
+      else if (this.editId === feature.properties._drawid) return;
+
+      this.editId = feature.properties._drawid;
+      coords = feature.geometry.coordinates;
+      var type = feature.geometry.type;
+
+      if (type === 'Point')
```
tmcw: Could we remove this mismatch between featureType and GeoJSON features and just have them be equal?

kelvinabrokwa: so get rid of the `square` type?

tmcw: Yes: note that in Leaflet-draw the 'square' draw tool is just a tuned version of the Polygon draw tool. It draws a rectangular polygon and once it's done drawing it, editing that shape is the same as editing any other polygon. Rectangles ⊂ Polygons

kelvinabrokwa: > once it's done drawing it, editing that shape is the same as editing any other polygon

hmm, unless I'm misunderstanding what you mean, this doesn't seem to be the case.

square

src/draw.js
```diff
@@ -91,47 +108,136 @@ Draw.prototype = extend(Control, {
...
+    this._map.getContainer().addEventListener('mousemove', this.drag, true);
+    this._map.getContainer().addEventListener('mouseup', this.endDrag, true);
+  },
+
+  _drag(e) {
+    e.stopPropagation();
+    if (!this.dragging) {
+      this.dragging = true;
+      this.init = DOM.mousePos(e, this._map.getContainer());
+      var options = R.merge(this.options, { geoJSON: this.editStore });
```
tristen: Is `R.merge` not the same as the [`xtend`](https://www.npmjs.com/package/xtend) that's currently in place?

kelvinabrokwa: oops, maybe, idk. I used `R.merge` here to replace that key in the object with this new value.

src/draw.js
tmcw: Err weird, misremembered.

tristen:
Awesome! Couple notes:

- A feature should remain in edit mode until the map is clicked or another feature is selected. Currently it moves out of edit mode after you drag it.
- Change the cursor from `crosshair` to `move` instead.
- (Understanding this will all change ...) The edit theme shouldn't turn a feature to a dash array but an edit mode color for now.

src/draw.js
tristen: Yeah, I think we can drop the `ramda` dep all together.

kelvinabrokwa:
To do before merge:

- [ ] Fix exit edit after drag bug
- [ ] Change cursor from `crosshair` to `move`
- [ ] Change edit mode color
- [ ] Remove ramda dependency

kelvinabrokwa:
I'm going to go ahead and merge this before this PR becomes anymore monstrous.

There is a bug with dragging line strings and at this point in time I'm convinced that it has to do with `map.featuresAt` which seems to have a number of [reported bugs][1]. There is also a small bug with dragging polygons (not squares). I will open issues about both of these and hit them tomorrow.

[1]: https://github.com/mapbox/mapbox-gl-js/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+featuresat

PR #91, opened 2015-10-15
Performance Review and Improvement

kelvinabrokwa:


kelvinabrokwa:
I don't know what this `(program)` deal in Chrome is but its the cause of the bad performance. All other browsers are quite smooth as it stands.

PR #99, opened 2015-10-27
Integrate Dataset API into example application

kelvinabrokwa:


kelvinabrokwa:
this is happening for real

PR #113, opened 2015-12-09
Permanent

drewbo:
The test is slightly long winded

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -14,18 +14,19 @@ export default class API extends mapboxgl.Control {
   /**
    * add a geometry
    * @param {Object} feature - GeoJSON feature
+   * @param {Object} options - add options to the geometry
    * @returns {Draw} this
    */
-  set(feature) {
+  set(feature, options) {
     feature = JSON.parse(JSON.stringify(feature));
     var id;
     if (feature.type === 'FeatureCollection') {
       id = [];
       for (var i = 0; i < feature.features.length; i++) {
-        id.push(this._setFeature(feature.features[i]));
+        id.push(this._setFeature(feature.features[i], options));
       }
     } else {
-      id = this._setFeature(feature);
+      id = this._setFeature(feature, options);
     }
     this._store._render();
     return id;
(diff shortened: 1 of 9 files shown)


drewbo:
@kelvinabrokwa this still has conflicts but let me know if this looks good otherwise and I can grab the updates from master then squash my commits.

src/api.js
```diff
@@ -55,6 +56,7 @@ export default class API extends mapboxgl.Control {
         console.log('MapboxGL Draw: Unsupported geometry type "' + feature.geometry.type + '"');
         return;
     }
+    feature.geojson.properties.permanent = permanent;
```
kelvinabrokwa: I don't think we should be changing (or adding to) the user's data's metadata. Perhaps a feature can be marked permanent in a geometry class field rather than directly in the GeoJSON properties.

Basically, if you load a bunch of features into gl-draw for editing it shouldn't come out with a bunch of random extra properties.

src/draw.js
```diff
@@ -220,7 +220,9 @@ export default class Draw extends API {
       if (err) throw err;
       if (features.length) { // clicked on a feature
         if (this._drawing) return;
-        this._edit(features[0].properties.drawId);
+        // check if the object is permanent
+        if (!features[0].properties.permanent)
```
kelvinabrokwa: This will change when you make the above adjustment.

drewbo:
@kelvinabrokwa really good suggestions. I updated the API to send an options object (which only supports one option right now: 'permanent') which is set on geometries and can be retrieved by calling the method `getOptions`.

kelvinabrokwa:
@drewbo looks good! After this is merged I'll probably join the other constructor parameters to the one options parameter. It will make adding features like this much easier and also simplify the code a bit.

drewbo:
Sweet. You want the commits squashed?

kelvinabrokwa:
@drewbo sure :)

kelvinabrokwa:
@drewbo good to go?

drewbo:
:+1:

Issue #112, opened 2015-12-08
Custom layer styles


drewbo:
Users should probably be able to pass options to change the default styles (it can be done now but after instantiation).

I'm going to need this for a project as well so I can take a first pass at this soon.

kelvinabrokwa:
@drewbo go for it! I've been meaning to start sketchy this for a while actually...

PR #114, opened 2015-12-09
Custom styles

drewbo:
It works! But definitely open to changes in both the code and the verbose documentation. Let me know what you think @kelvinabrokwa 

Re: #112

screen shot 2015-12-09 at 4 11 53 pm


API.md
```diff
@@ -27,6 +27,29 @@ drawing | boolean | The ability to draw and delete features - default: `true`
 interactive | boolean | Keep all features permanently in edit mode - default: `false`
 keybindings | boolean | Keyboard shortcuts for drawing - default: `true`
 controls | Object | drawable shapes - default `{ marker: true, line: true, shape: true, square: true }`
+styles | Object | Add a style with any of the following properties: `gl-draw-polygon`, `gl-draw-polygon-stroke`, `gl-draw-line`, `gl-draw-point`, `gl-drawing-line`, `gl-drawing-polygon`, `gl-drawing-polygon-stroke`, `gl-drawing-points`, `gl-edit-line`, `gl-edit-polygon`, `gl-edit-polygon-stroke`, `gl-edit-point`, `gl-edit-point-mid`. The property should be an object with either the `layout` and/or `paint` properties as specified in the [Mapbox GL Style Reference](https://www.mapbox.com/mapbox-gl-style-spec/). It will overwrite the corresponding default  styles found in [`src/theme/`](https://github.com/mapbox/mapbox-gl-draw/tree/master/src/theme).
```
kelvinabrokwa: can you please make the properties into a bulleted list for the sake of readability?

kelvinabrokwa:
@drewbo nice work! do you think its gl-draw's responsibility to validate the user's styles or should be just let gl-js throw on invalid styles?

I'd say option 2.

API.md
drewbo: You don't like the list of 13 property names in a row?! (Will do)

drewbo:
Thanks! Agree on option 2. This could also break other things if a user sends a style which has a `filters` or `source` property. Do we want to be "defensive" and prevent that?

kelvinabrokwa:
@drewbo we can worry about that later. This is a great feature.

drewbo:
Cool. This doesn't have any associated tests. Does it need anything? Otherwise I will squash the commits again.

kelvinabrokwa:
@drewbo nah just the docs clean up and then I'll merge.

drewbo:
Cool, I just hit the properties list thing

kelvinabrokwa:
@drewbo duuuuudeeee, sorry to be so demanding, but markdown, please...

drewbo:
I don't think there is a way to write markdown lists inside a table? If there is, happy to reformat

kelvinabrokwa:
@drewbo I believe you. Thanks for the PR! Merging

PR #121, opened 2015-12-10
move mapbox-gl to a peer dependency

### LINKED ISSUE #120: mapbox-gl as a peer dependency?
Having mapbox-gl as a dependency bloats build files. The directions control solves this problem by [including mapbox-gl as a peer dependency](https://github.com/mapbox/mapbox-gl-directions/blob/9d4745763649db104c539e7b04ea493d3b2f0af9/package.json#L52-L54). Any reason we shouldn't do that here?

/cc @kelvinabrokwa @tristen @scothis @tmcw

mcwhittemore:
Closes #120

@scothis for the review

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,11 +44,14 @@
     "tap-closer": "^1.0.0",
     "tap-status": "^1.0.1",
     "tape": "^4.0.0",
-    "uglify-js": "^2.4.23"
+    "uglify-js": "^2.4.23",
+    "mapbox-gl": "^0.12.0"
+  },
+  "peerDependencies": {
+    "mapbox-gl": "^0.12.0"
   },
   "dependencies": {
     "hat": "0.0.3",
-    "mapbox-gl": "^0.11.0",
     "ramda": "^0.17.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,11 +44,14 @@
     "tap-closer": "^1.0.0",
     "tap-status": "^1.0.1",
     "tape": "^4.0.0",
-    "uglify-js": "^2.4.23"
+    "uglify-js": "^2.4.23",
+    "mapbox-gl": "^0.11.0"
+  },
+  "peerDependencies": {
+    "mapbox-gl": "^0.11.0"
```
scothis: 0.12.1 is the latest stable version of mapbox-gl, the peerDependency should include both 0.11.x and 0.12.x since both are API compatible.

You can either use `^0.11.0 || ^0.12.0` to allow both, or take the latest 0.12 and force gl-draw users to upgrade.

mcwhittemore:
Going to force them to upgrade, this was just released yesterday and its at 0.0.0

PR #122, opened 2015-12-10
use featuresIn to select features for editing

mcwhittemore:
This decreases the size of bundled versions of mapbox-gl-draw by removing turf-intersect and opting for the maps default featureIn selector.

@kelvinabrokwa or @tmcw for the review


kelvinabrokwa:
@mcwhittemore lgtm

PR #125, opened 2015-12-11
Fix 124

### LINKED ISSUE #124: using set directly puts all coordinates to zero
`.set` puts the feature information on [`options.feature`](https://github.com/mapbox/mapbox-gl-draw/commit/a154758695dd1b4f44f7aa6acde24602e7bfdf76#diff-e1d62d8c4965edf0e532dfa19da0891fR54) but then reads it from [`options.data`](https://github.com/mapbox/mapbox-gl-draw/commit/a154758695dd1b4f44f7aa6acde24602e7bfdf76#diff-a1a6578e756b54c1dfad065d5347c4e8R20) (which doesn't exist) so all coordinates are set to [0,0].

This doesn't get caught by tests because the test point is the same as the default coordinates

drewbo:
@kelvinabrokwa resolves #124, updated the test to catch cases like this in the future

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -51,7 +51,7 @@ export default class API extends mapboxgl.Control {
       options = {};
     }
     options.map = this._map;
-    options.feature = feature;
+    options.data = feature;
 
     switch (feature.geometry.type) {
       case 'Point':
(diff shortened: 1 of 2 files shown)


kelvinabrokwa:
@drewbo thanks dude!

Issue #129, opened 2015-12-11
Slim down utils


tmcw:
There are a few utils, like DOM.destroy, which are relatively unnecessary shortcuts and should be replaced by the literal code.

* no dom-destroy https://github.com/mapbox/mapbox-gl-draw/pull/131

PR #130, opened 2015-12-11
Remove DOM.destroy util

tmcw:
Refs #129


tmcw:
Had an old version of `master`, nevahmind

PR #141, opened 2015-12-14
Hide internal events from end user and standardize external events

### LINKED ISSUE #138: External vs internal events
@kelvinabrokwa I've been playing with gl-draw a bit today. For a project I want to take features out of draw when the `I'm done drawing` event is fired. That said, this event is also used by gl-draw so it is possible that I'd start listening before the inner workings of gl-draw set up their own listeners. If the actions gl-draw takes on those actions are important to me, I'm out of luck. I'd like to look into standardizing some of these events and create a good separation between events for end users and triggers to needed for the inner workings of gl-draw. What do you think?

mcwhittemore:
Currently gl-draw uses the public mapbox-gl-js event emitter for its own internal state management. This has the nice benefit of giving end users a way to connect into the gl-draw life cycle, but it also makes it hard to create good internal updates as external and internal api contacts are tightly bound. This PR separates these concerns and starts to set standards around external event naming and payload content. 

**This is a breaking change.** As we are still in 0.*.* range, this will not mean a major bump, but I want to make sure anyone who is watching  this today is aware of the direction @kelvinabrokwa and I are looking to go. I also didn't recreate all of `0.0.2`s external events. If there are no major complaints, my goal is to merge this tomorrow as `0.1.0` so that pacakge.json files that used the ^ operator.

Closes #138 

/cc @drewbo @tmcw @edenh

### DIFF
--- a/src/internal_events.js
+++ b/src/internal_events.js
@@ -1,0 +1,5 @@
+// TODO: Remove the need for internal event handling
+// by more clearly and explicitly defining the relationship
+// between geometries, draw.js and the stores.
+
+export default new (require('events').EventEmitter);
--- a/src/store.js
+++ b/src/store.js
@@ -1,5 +1,7 @@
 'use strict';
 
+import InternalEvents from './internal_events';
+
 /**
  * A store for keeping track of versions of drawings
  *
@@ -12,7 +14,7 @@ export default class Store {
     this._map = map;
     this._features = {};
     this._editStore = null;
-    this._map.on('drawing.end', e => {
+    InternalEvents.on('drawing.end', e => {
       this.set(e.geometry);
     });
   }
@@ -49,6 +51,12 @@ export default class Store {
    */
   set(feature, preventRender) {
     this._features[feature.drawId] = feature;
+
+    this._map.fire('draw.set', {
+      id: feature.drawId,
+      geojson: feature.geojson
+    });
+
     if (!preventRender) {
       this._render();
     }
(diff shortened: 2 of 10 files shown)


src/internal_events.js
```diff
@@ -0,0 +1,10 @@
+// TODO: Remove the need for internal event handling
+// by more clearly and explicitily defining the relationship
```
tmcw: sp

src/store.js
```diff
@@ -48,7 +50,14 @@ export default class Store {
    * @param {Object} feature - GeoJSON feature
    */
   set(feature, preventRender) {
+    console.log('set', feature);
```
tmcw: delete

src/internal_events.js
```diff
@@ -0,0 +1,10 @@
...
+// between geometries, draw.js and the stores.
+
+var emitter = require('event-emitter')({});
+
+export default {
```
tmcw: why can't this do

```
export default emitter;
```

?

package.json
```diff
@@ -51,6 +51,7 @@
     "mapbox-gl": "^0.12.0"
   },
   "dependencies": {
+    "event-emitter": "^0.3.4",
```
tmcw: could we use [events](https://www.npmjs.com/package/events) such as to be the same api as node?

tmcw: any by that, i mean `require('events')` without requiring a new module, which should work cc @tristen

src/internal_events.js
mcwhittemore: needs to map to fire

package.json
mcwhittemore: Oh! Nice! My least favorite thing here was adding the dep.

src/internal_events.js
mcwhittemore: oh, this dumb, I was thinking about mapbox-gl. GP.

Issue #94, opened 2015-10-22
Drag performance in Chrome


kelvinabrokwa:
What is this `(program)` business?

![](https://cloud.githubusercontent.com/assets/7795202/10674766/c15494bc-78ca-11e5-8b2d-ccda966ecddd.png)

mh166:
Maybe this SO answer can help you a bit: [Slow javascript execution in chrome, profiler yields (program)](http://stackoverflow.com/a/9574976/3380671).

PR #147, opened 2015-12-16
Common Interface for sub controls and one store

mcwhittemore:
@edenh & @kelvinabrokwa 

Here is where I am in my pursuit of moving gl-draw to enforce a standard interface for its sub controls and then move it to one store. There are few rendering bugs that I'm working on currently. Once I've got those fixed, the common interfaces need to be touched up a bit and documented and than I'll move onto the single store.

There are a few things I'm not too happy about (draw_events for one) so I might try to clean those up a bit more too.

I'd love some feedback and questions to what I'm doing here to make sure I'm not missing something major.


mcwhittemore:
Currently this is at 48-52 FPS for a few features.

## Bugs fixed by this

- [ ] #148

mcwhittemore:
The last thing that needs to go on here is to go through the api docs and make sure nothing was broken. I think we should just dump these use cases into debug for now and then write them as tests after this lands.

@kelvinabrokwa thoughts?

@tmcw wanna take sometime to review this?

kelvinabrokwa:
@mcwhittemore 
TO DO

- [ ] trash button doesn't show on shift-drag selection

mcwhittemore:
@kelvinabrokwa wanna take a shot at those two things?

kelvinabrokwa:
@mcwhittemore yup

mourner:
@mcwhittemore this is really cool! What was the main performance hog for #94?

mcwhittemore:
@mourner - there were a few things I did in general. One was to remove the geometry request in the featureAt calls. The second was to reduce the featureAt calls and the setData calls. Now we are only setting data once or twice per frame where before we were setting in on every mouse event.

mcwhittemore:
To be honest, I'm still not sure how much of a perf gain this branch has. Sometimes I see a 10fps gain, other times I see a 1 to 2 fps drop.

PR #156, opened 2015-12-21
incorrect example urls in readme

twelch:
example folder names changed looks like


mcwhittemore:
Thanks for catching this!

Issue #157, opened 2015-12-21
Great Drawing Tool or Great setData abstraction api?


mcwhittemore:
@edenh and I have been talking a bit over the last few days about what gl-draw should be. On one hand gl-draw could be a fully featured drawing tool, on the other it could be a set of api's to make working with the geojson and setData easier.

Currently gl-draw has been moving towards option one with a few option two features. But as we start to use it more and more questions about how custom one would want to make their drawing tools lead us consider option two. 

We could go both routes, but I think that would lead to a rather bloated api.

A few issues that touch on this tension: #154, #109, #139

/cc @kelvinabrokwa @drewbo @tmcw @musicformellons @samanpwbb

samanpwbb:
> a set of api's to make working with the geojson and setData easier.

I'd be great to have this. Will open up a wider range of use cases for the project and give users more flexibility to build exactly the kinds of draw interactions that want. 

We wouldn't need to remove the existing UI components – they could stay in the project and use the APIS. Developers would then have the choice to directly access the API and build their own UI or use the existing UI components.

drewbo:
My use case so far has been much closer to the second option than the first (I removed the UI elements but leave in the ability to draw a polygon). I like @samanpwbb's suggestion, basically it ends up being both options but is very light on the drawing specific features. No strong feelings either way from me.

mcwhittemore:
Yea, most projects are going to create their own buttons to start drawing. We will need an api to start drawing these features.

I'm still concerned with what a more exhaustive api approach means. I'll post in a bit when my thoughts are better formed.

mcwhittemore:
I got my thoughts down in #158. After that, @edenh and I chatted a bit and realized that dropping the idea of editing from gl-draw and moving to selected helps clear up a bunch of my concerns. Do this would push us more towards @samanpwbb suggestion.

Issue #154, opened 2015-12-18
Methods for manipulating edit state


edenh:
`Draw.setEdit(String: drawId)`: select a single feature as editable
`Draw.unsetEdit(String: drawId)`:  deselect a single feature
`Draw.setEditAll()`: select all features
`Draw.unsetEditAll()`: deselect all features

mcwhittemore:
@edenh - what is the goal of this api? When would you want to programmatically start/stop editing?

mcwhittemore:
From another ticket I see you want to have selecting done via key bindings. Currently gl-draw does its own key bindings, I think we should keep it that way.

PR #158, opened 2015-12-21
Add selection methods, tests, and cleanup

edenh:
Ref #154, #157 

- adds `Draw.select`, `Draw.selectAll`, `Draw.deselect`, and `Draw.deselectAll`
- renames `Draw.set` to `Draw.add`
- tests
- whitespace cleanup


mcwhittemore:
#157 is maybe too philosophical, so I'm going to use this PR to help bring my thoughts down a bit.

`Draw.select` and `Draw.selectAll` moves features from the non editing to the editing state. This is currently done by gl-draw when a users clicks on the feature. Key bindings being part of gl-draw already, you can assume `cmd+a` would perform a `Draw.selectAll()` internally.

While these two use cases are the obvious ones, you could also imagine an edit button in an app that would use `Draw.select` to start editing one feature.

You can also imagine a user wanting to highlight a set of features and using a bunch of `Draw.select` calls to put the features into edit mode which changes their style.

For `Draw.unselect` you can imagine a user wanting to stop editing when their user clicks on a different tab in their application.

One problem I see here is that this muddles the concept of selected with editable. I think we will want to add vertex selection in time. This would allow users to grab a few vertices and move them at the same time. Maybe this is something we should ignore until we get there? Worse case it would be a breaking change...

Also, the use case of using selectAll to show highlight a feature by putting it into edit mode, sounds like a request to style features using more than the two default buckets of gl-draw.

All of that said, we're still in 0.* so lets just do stuff and refactor later when we have more than theoretical other options.

mcwhittemore:
We should introduce a changelog and mention that this is a breaking change.

src/api.js
```diff
@@ -106,15 +106,57 @@ export default class API extends mapboxgl.Control {
...
+    return this;
+  }
+
+  /**
+   * deselect all features for editing
+   * @param {String} id - the drawId of the feature
+   * @returns {Draw} this
+   */
+  deselectAll() {
+    this._store.getAllIds()
```
mcwhittemore: Lets filter this first, just in case the unneeded calls to commit make more renders and slow things down.

src/api.js
```diff
@@ -106,15 +106,57 @@ export default class API extends mapboxgl.Control {
...
+      .forEach(id => this.commit(id));
+    return this;
+  }
+
+  /**
+   * get a feature collection of features being edited
    * @return {Object} a feature collection of geometries that are in edit mode
    */
-  getEditing() {
+  getSelected() {
```
mcwhittemore: What are you using this for? Seems weird to me to change the api. Are you looking for something to give you the list of feature ids that are in edit mode?

src/store.js
```diff
@@ -86,21 +86,21 @@ export default class Store {
    */
   clearEditing() {
```
mcwhittemore: This should be changed to delete too.

src/api.js
mcwhittemore: `_store.getEditIds.forEach` would be better even

mcwhittemore:
@edenh - this is looking good. I think we should push this a bit more and just get rid of the concept of editing. This would help kill some of my fears from the long post above and I think would help make the code a bit cleaner.

src/api.js
edenh: I think it should commit. UX-wise, I would assume deselecting does not "revert" my changes.

edenh:
Updated to use `selected` terminology over `edit`, as well as a changelog.

mcwhittemore:
Looks good, but I don't see the changelog

edenh:
Added here https://github.com/mapbox/mapbox-gl-draw/commit/6827ebe3f512c680234006fde4ec649623fba825

src/api.js
```diff
@@ -13,14 +13,14 @@ export default class API extends mapboxgl.Control {
...
-   * @pata {Boolean} [options.permanent=false] - disable editing for feature
+   * @pata {Boolean} [options.permanent=false] - disable selection for feature
    * @returns {Number} draw id of the set feature or an array of
    * draw ids if a feature collection was added
    */
-  set(feature, options) {
+  add(feature, options) {
     feature = JSON.parse(JSON.stringify(feature));
     if (feature.type === 'FeatureCollection') {
       return feature.features.map(subFeature => this.set(subFeature, options));
```
kelvinabrokwa: this should also be changed to `this.add`

kelvinabrokwa:
we should also be firing `draw.add` if we've changed the naming scheme

https://github.com/mapbox/mapbox-gl-draw/blob/master/src/store.js#L32-L35
https://github.com/mapbox/mapbox-gl-draw/blob/master/src/store.js#L132-L135

PR #167, opened 2016-01-05
Set drawId on .add via the feature id property

### LINKED ISSUE #136: set drawId when adding a feature
Currently `.set` doesn't have anyway to specify an id. To help users avoid needed to map their ids to draw's ids, it would be nice to be able to pass in an id. Maybe with a `.put` function.

mcwhittemore:
Closes #136


mcwhittemore:
For @edenh to review

edenh:
:+1: lgtm

PR #171, opened 2016-01-06
set feature ready for rendering on add

mcwhittemore:
Currently they won't render because they are set to ready = false.


mcwhittemore:
@kelvinabrokwa for the review

kelvinabrokwa:
@mcwhittemore :shipit:

Issue #153, opened 2015-12-18
Change Draw.remove to Draw.destroy


kelvinabrokwa:
As a MapboxGL.JS control, the `remove` method should remove it from the map (it shouldn't be used for removing features).

kelvinabrokwa:
This will be a breaking change

mcwhittemore:
+1

PR #175, opened 2016-01-13
add removal and change api `remove` to `destroy`

kelvinabrokwa:
This change, as is, only works when render calls are not debounced. However, without debouncing, performance is worse.
I will continue to explore alternatives.

[EDIT]: This has nothing to do with debouncing
This should unblock tests.


mcwhittemore:
@kelvinabrokwa - why do you want to change remove to destroy?

kelvinabrokwa:
@mcwhittemore https://github.com/mapbox/mapbox-gl-draw/issues/153

kelvinabrokwa:
@mcwhittemore :eyes: plz

src/api.js
```diff
@@ -185,4 +185,10 @@ export default class API extends mapboxgl.Control {
     return this;
   }
 
+  remove() {
```
mcwhittemore: Lets move this to draw.js, as its being used by mapbox-gl

kelvinabrokwa: done

mcwhittemore:
@kelvinabrokwa - I'm getting an error when trying to load debug.

image

kelvinabrokwa:
@mcwhittemore yeah... so it turns out there is no ~public~ way to know if `load` has fired and since draw depends on that to set styles and sources it is better to `map.addControl` in the user's `map.on('load')` callback. I need to document this.

mcwhittemore:
@kelvinabrokwa can you update debug to not fail and document what needs to be documented?

kelvinabrokwa:
@mcwhittemore debug update as of fd1ffad and docs coming up!

kelvinabrokwa:
@mcwhittemore https://github.com/mapbox/mapbox-gl-draw/blob/068235558bbc0eec781b50df47c97409ca7c5ce9/src/draw.js#L73 is a bit of a hack but it lets you safely add draw to the map before load fires. The upside is better backwards compatibility (and no need for new docs)

mcwhittemore:
Merge, version, push and publish!

PR #204, opened 2016-02-02
add startDrawing to public API [WIP]

### LINKED ISSUE #196: Expose startDrawing to the public api
Users should be able to opt out of our controls and still start drawing. This would be possible via a public version of `_startDrawing`. I think this should accept constant strings which we expose on Draw. Thoughts @kelvinabrokwa?

### LINKED ISSUE #216: Clicking on a draw button again before drawing anything will commit a premature feature
If you click on any of the draw buttons and then, instead of drawing anything, you click on a draw button again - a feature will now have been committed with a geometry with 0-coordinates. If you continue clicking on the draw buttons you will continue adding premature features.

kelvinabrokwa:
Closes #196
Closes #216

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -111,12 +111,7 @@ export default class API extends mapboxgl.Control {
       });
   }
 
-  startDrawing(type) {
-    this._startDrawing(type);
-  }
-
-  _startDrawing() {
-    // implemented in draw.js
+  startDrawing() {
     return null;
   }
 
--- a/src/api.js
+++ b/src/api.js
@@ -10,6 +10,12 @@ export default class API extends mapboxgl.Control {
 
   constructor() {
     super();
+    this.types = {
+      SQUARE: 'square',
+      POLYGON: 'polygon',
+      LINE: 'line',
+      POINT: 'point'
+    };
   }
 
   /**
@@ -103,6 +109,15 @@ export default class API extends mapboxgl.Control {
         type: 'FeatureCollection',
         features: []
       });
+  }
+
+  startDrawing(type) {
+    this._startDrawing(type);
+  }
+
+  _startDrawing() {
+    // implemented in draw.js
+    return null;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/api.js
```diff
@@ -10,6 +10,12 @@ export default class API extends mapboxgl.Control {
 
   constructor() {
     super();
+    this.types = {
```
mcwhittemore: Should these be on this or should we have a constants object on the Draw constructor?

mcwhittemore: I don't know, many constants are a bit weird. People will likely just enter their own strings...

src/api.js
```diff
@@ -105,6 +111,15 @@ export default class API extends mapboxgl.Control {
       });
   }
 
+  startDrawing(type) {
+    this._startDrawing(type);
+  }
+
+  _startDrawing() {
```
mcwhittemore: Lets drop private functions. What do we gain by making this kinda private?

src/api.js
kelvinabrokwa: Constants will help us prevent bugs internally and externally. And I put them in API.js since its public-facing.

mcwhittemore: Sounds good.

mcwhittemore:
Lets merge

kelvinabrokwa:
@mcwhittemore :shipit:?

src/draw.js
```diff
@@ -151,20 +151,20 @@ export default class Draw extends API {
...
         break;
-      case 'square':
+      case this.types.SQUARE:
         obj = new Square({ map: this._map });
         break;
-      case 'point':
+      case this.types.POINT:
         obj = new Point({ map: this._map });
         break;
       default:
```
mcwhittemore: Should this through an error?

mcwhittemore:
- [ ] add debug implementation (more buttons that start the drawing)
- [ ] Can we test this?

mcwhittemore:
Ok. This LGTM. We still need to find write tests, but I think we should do this after we land a way to test draw without using gl. I'm pretty sure we can do this with a simple shim of mapbox-gl-js.

PR #209, opened 2016-02-05
Don't show trashcan when nothing is selected - fixes #166

### LINKED ISSUE #166: map zoom in show trash button
trash button show when zoom in the map

load map:
image

zoom in map:

image

mcwhittemore:
@rbonillajr or @averas can one of you confirm that this fixes the map zoom trash can error from #166?

### DIFF
--- a/src/draw.js
+++ b/src/draw.js
@@ -65,7 +65,7 @@ export default class Draw extends API {
    */
   onAdd(map) {
     var container = this._container = DOM.create('div', 'mapboxgl-ctrl-group', map.getContainer());
-    this._store = new Store(map);
+    this._store = new Store(map, this);
 
     this._map = map;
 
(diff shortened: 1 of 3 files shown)


averas:
Yes! Trashcan does not appear as a result of plain shift-bbox-zoom anymore.

A sidenote, however, is that when using shift-bbox-zoom over editable features both a bbox-zoom _and_ selection of the features are triggered.

Not sure what would be the best UX here.. not zooming in the case features are selected would probably be unintuitive (and probably a bit cumbersome to implement). Not being able to select multiple features at all would certainly be a loss of a valuable feature. Perhaps a different key? CTRL-bbox for selection?

mcwhittemore:
Thanks for confirmation @averas.

I agree the UX isn't right here. Wanna open another ticket where we can talk about that?

PR #211, opened 2016-02-05
Fix end of drawing

mcwhittemore:
- [x] Double click on polygon after putting down a single point shouldn't call set


mcwhittemore:
@lucaswoj or @kelvinabrokwa wanna take a look at this?

PR #222, opened 2016-02-09
Expand mapbox-gl version range to include 0.14.0

scothis:


scothis:
There is one breaking change in 0.14:
- Switch GeoJSONSource clustering options from being measured in extent-units to pixels (#2026)

scothis:
@mcwhittemore for review

PR #229, opened 2016-02-10
use Object.assign with defaultControls instead of several ternaries

conorhastings:


mcwhittemore:
@conorhastings - I kept tending towards trying to write a deep merge and knew that was too much so I went with the verbose copy. This is much better then that. Thanks.

Issue #215, opened 2016-02-08
Addition and removal of sources and layers to the map should be done via the batch API


averas:
Adding and removing sources and layers one-by-one takes unnecessary long time. They should be wrapped in the batch API. I can provide a PR if you agree.

kelvinabrokwa:
@averas go for it :smile:

averas:
Fixed by #242.

PR #242, opened 2016-02-21
Wrapped layer/source add/removes in batch calls

averas:
Solves #215.

Brings down initialization of draw from ~700ms to ~100ms and removal from ~700ms to ~90ms, which makes a great deal of difference in applications where you enable/disable draw often.

cc @mcwhittemore @kelvinabrokwa

### DIFF
--- a/src/draw.js
+++ b/src/draw.js
@@ -248,8 +248,8 @@ export default class Draw extends API {
         Object.assign(style, self.options.styles[style.id] || {});
         batch.addLayer(style);
       }
+      self._store._render();
     });
-    self._store._render();
   }
 
   _removeLayers() {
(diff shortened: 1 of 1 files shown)


src/draw.js
```diff
@@ -217,52 +217,56 @@ export default class Draw extends API {
...
   _setStyles() {
-    // drawn features style
-    this._map.addSource('draw', {
-      data: {
-        type: 'FeatureCollection',
-        features: []
-      },
-      type: 'geojson'
-    });
+    var self = this;
```
kelvinabrokwa: I think using [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) as callbacks will bind `this` to the functions so that you don't have to alias `this` as `self`

averas:
@kelvinabrokwa You got it. :eyes:

kelvinabrokwa:
@averas great work!

PR #244, opened 2016-03-01
Moving to a mock for tests

### LINKED ISSUE #197: events.tests.js is very fickle
The draw.set seems to never work (though we use it in other places) and if we skip this tests, draw.delete often (but not always) doesn't run. We need to look into if this is a problem with `map.once`, the order we're loading in things, etc.

### LINKED ISSUE #200: Get tests running on circle-ci
We need to find a way to get as many tests as we can to run on circle. Currently we're just pretending to have run our tests and this leads us to a false sense of security.

### LINKED ISSUE #235: Remove direct mapbox-gl dependencies
Currently `mapbox-gl-draw` include `mapbox-gl` as part of its build. This means end users are downloading `mapbox-gl` twice. Lets find ways to remove these dependencies.

- [ ] https://github.com/mapbox/mapbox-gl-draw/blob/33346ac48d33480eec3ef76400b1b82aa0a09e2e/src/api.js#L3
- [ ] https://github.com/mapbox/mapbox-gl-draw/blob/85141d0c812862c92398e2cf051eddab09abeb79/src/util.js#L3

mcwhittemore:
Closes #235
Closes #200
Closes #197

This also makes it so `permanent` features aren't selectable with `Draw.select()`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
     ]
   },
   "scripts": {
-    "test": "npm run lint && npm run prepublish",
-    "test-unit": "browserify test/*.test.js | tap-closer | smokestack | tap-status",
+    "test": "npm run lint && tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
     "lint": "eslint --no-eslintrc -c .eslintrc index.js src",
     "prepublish": "NODE_ENV=production browserify index.js | uglifyjs -c -m > dist/mapbox-gl-draw.js",
     "start": "budo index.js --serve=dist/mapbox-gl-draw.js --live -d | bistre"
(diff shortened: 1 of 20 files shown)


mcwhittemore:
Can someone take a look at this and make sure all is well?

/cc @kelvinabrokwa @drewbo @averas

package.json
```diff
@@ -12,8 +12,7 @@
     ]
   },
   "scripts": {
-    "test": "npm run lint && npm run prepublish",
-    "test-unit": "browserify test/*.test.js | tap-closer | smokestack | tap-status",
+    "test": "tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
```
drewbo: Is the removal of linting here intentional?

src/feature_types/square.js
```diff
@@ -32,12 +32,14 @@ export default class Square extends Polygon {
   }
 
   onClick(e) {
+    console.log('clicking', this.ready);
```
drewbo: Probably should remove the `console.log`s (there are a few)

package.json
mcwhittemore: Yes. The test-unit used to be the real tests but they couldn't run in circle so for circle we just tested if we could build. I don't think we need to test this anymore, but if you have concerns about not testing the build in circle (its implicitly tested by publishing) I'm all ears.

drewbo: It's nice for making sure that PRs don't introduce linting problems to fail the tests on this

kelvinabrokwa: +1 on including `npm run lint`

src/draw.js
```diff
@@ -83,6 +83,7 @@ export default class Draw extends API {
       this._setStyles();
     } else {
       map.on('load', () => {
+        console.log('loading');
```
kelvinabrokwa: rm `console.log`

src/feature_types/feature.js
```diff
@@ -64,6 +64,7 @@ export default class Geometry {
   */
 
   onStopDrawing() {
+    console.log('super.onStopDrawing');
```
kelvinabrokwa: ^^

drewbo:
@mcwhittemore LGTM other than the things noted. Should probably try to add back in the skipped tests at some point but not necessary for this PR

PR #246, opened 2016-03-02
Update mapbox-gl peerDependency

drewbo:
Maybe a little overkill at this point but don't want to leave behind `0.13` yet

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "uglify-js": "^2.4.23"
   },
   "peerDependencies": {
-    "mapbox-gl": "^0.13.1 || ^0.14.0"
+    "mapbox-gl": "^0.13.1 || ^0.14.0 || ^0.15.0"
   },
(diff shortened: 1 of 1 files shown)


lucaswoj:
My understanding is that `^0.13.0` includes `0.13.x` as well as `0.14.x`, `0.15.x`, etc.

`~0.13.0` would only include `0.13.x`

We should remove the `|| ^0.15.0`

cc @kelvinabrokwa @mcwhittemore

drewbo:
I think pre `1.0.0` npm assumes that minor versions can be breaking. Maybe it could update to `<=` syntax?

https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

lucaswoj:
Ah! You're right @drewbo.

mcwhittemore:
This looks good. I'll push out a release.

PR #251, opened 2016-03-07
better namespacing, closes #116

### LINKED ISSUE #116: CSS is not namespaced
`.trash`, [for example](https://github.com/mapbox/mapbox-gl-draw/blob/9323dbdb0ab85e733b0e3899a67b5a207fc43036/dist/mapbox-gl-draw.css#L49), overrides the global `.trash` rather than creating a mapbox-gl-draw trash style.

mayagao:
ref https://github.com/mapbox/mapbox-gl-draw/issues/116
cc @mcwhittemore


dist/mapbox-gl-draw.css
```diff
@@ -34,21 +34,21 @@
...
 }
-.trash {
+.mapboxgl-draw_trash {
```
mcwhittemore: @mayagao why `mapboxgl-draw_trash` and not `mapboxgl-draw-trash`?

mcwhittemore:
Code wise this looks good to me. I need to play with it before we merge. Also, this is a breaking change so we will need to bump semvar minor.

dist/mapbox-gl-draw.css
lucaswoj: :+1: to `mapboxgl-draw-trash`

mayagao:
@mcwhittemore I'm following @samanpwbb 's rules here : 
> - Every rule should be exactly one class unless absolutely necessary.
> - Prefix all rule names with an exact copy of the project name in npm
> - Descriptive parts of the rule should follow after a single or double _

PR #252, opened 2016-03-07
add options to disable and enable all controls, closes #231

### LINKED ISSUE #231: Find a clear way to disable all controls all the time
Currently `mapbox-gl-draw` allows a user to pass `drawing: false` to the config object to disable all controls. We should find a better way to do this and deprecate `drawing: false` as drawing indicates that this stops the ability to draw, not the disabling of all controls.

One option is that we could expand the controls attribute on the user provided config to be either false, an objective or undefined. Where false turns off all controls, an object lets a user turn of some controls and undefined uses the defaults (all controls on).

Another option is that we could let the config look something like: `{controls: { off: true }}`.

@drewbo I know you are using this feature. Any suggestions to the best api?

mayagao:
Closes #231


mcwhittemore:
@mayagao this is looking good to me.

@drewbo or @timjcook I think this covers what we talked about in #231. Can you confirm?

drewbo:
:+1:

timjcook:
:ok_hand:

PR #255, opened 2016-03-09
fix bug where ui doesn't update on update

### LINKED ISSUE #233: Draw.update should re-render the view

mcwhittemore:
Closes #233


mcwhittemore:
I'd really love tests for this, but I'm not sure if we can do that currently. Maybe render should emit something so we can listen for that?

PR #258, opened 2016-03-14
0.6.0-rc1

### LINKED ISSUE #109: Create a mode that doesn't let the user edit the map
Would be nice to have support for objects which can't be edited once added (either just by disabling click editing or disabling future `.update` calls).

Boolean could be set either on the GeoJSON properties or passed directly to `.set`.

I started a quick & dirty version at https://github.com/drewbo/gl-draw/tree/permanent for my own purposes but can work on it further (docs + tests) if this is a desired feature.

### LINKED ISSUE #128: Winding order matters for polygons and cuts out other features
winding

Winding order shouldn't matter

### LINKED ISSUE #218: Event for when vertices are added/modified on LineString and Polygon
I have a use case where drawing is limited to certain areas. For the Point type I immediately destroy the feature when created if it's out of bounds. For LineString and Polygon features I currently wait until it is committed before I can do any validation, which UX suffers from. I would love to see modification events both on committed and non-committed features so I can carry out validation on its current properties, immediately removing any vertices that are out of bounds.

### LINKED ISSUE #243: Create performance benchmarks
We should create some performance benchmarks for mapbox-gl-draw that are quantifiable and repeatable. 

Ideally, these benchmarks would be runnable in a headless environment using something like http://karma-runner.github.io. 

It probably makes sense to work on the upstream mapbox-gl-js performance benchmarks first, and use the lessons and infrastructure from those to build the mapbox-gl-draw benchmarks.

ref https://github.com/mapbox/mapbox-gl-js/issues/954
ref https://github.com/mapbox/mapbox-gl-js/issues/1362

### LINKED ISSUE #245: Every edit to a drawn feature should be emitted
Ref #218 for a simular request

### LINKED ISSUE #247: Make feature selection easier to do in the code
I want to add different cursors styles when I'm hovering over a feature's vertex/boundaries. Right now this is really hard to do, especially when two features overlap or two vertexes are near each other. It's not clear which one will be selected.

### LINKED ISSUE #249: 'interactive' option is broken
As it stated in documentation:

option | values | function
--- | --- | ---
interactive | boolean | Keep all features permanently in selected mode - default: `false`

But it doesn't work. When I create mapbox.Draw with `interactive: true` - I still can leave selected mode if I click somewhere outside of shape.

### LINKED ISSUE #257: Defer events, separate draw modes
https://github.com/mapbox/mapbox-gl-draw/blob/master/src/draw_events.js#L140

https://github.com/mapbox/mapbox-gl-draw/blob/master/src/draw_events.js#L5

An ideal architecture would let us separate draw modes so that the logic for each type of drawing is separate. Something like

* There's a global click handler that does featuresAt
* Each mode can subscribe to specific types of clicks and on specific features

mcwhittemore:
Closes #257
Closes #249 - selection is not longer a global concept in Draw so forcing feature to be selected is no longer viable.
Closes #247
Closes #245 - draw.changed is emitted with ever render
Closes #243
Closes #218 - draw.changed is emitted with ever render. Adding vertices causes a render.
Closes #128 - fixed in mabpbox-gl@0.16.0
Closes #109 - permanent features have been removed

---

This PR is moving us away from a tool where features describe how events affect them to a system where modes describe how they effect features. Currently you can draw a feature, select a feature and change a features coordinates at almost any time. This is communicated to the end-user via two styles (selected and unselected). Moving forward each of these actions will be available via different modes and only one mode will be running at a time.

## Overview of modes

Below is a list of the current modes and what they can do. I also need to document what styling options this gives the end-developer both in CSS and the map style.

The draw modes are always available via the draw controls or via `api.startDrawing`

I think all modes should be enterable via the api.

### Features Select

- Select a feature
- Multi select features
- Drag selected features
- Delete selected features
- Move to `Direct Select` by doubling clicking on a feature

### Direct Select

- Select a vertex
- Multi select vertices
- Move selected vertices
- Delete selected vertices
- Move to `Features Select` by double clicking any where

### Draw LineString

Lets you draw a LineString. Exits to `Features Select`

### Draw Polygon

Lets you draw a Polygon. Exits to `Features Select`

### Draw Point

Lets you place a point. Exits to `Features Select`

## Notes

This is still a long way out

The goals of this PR are:
- rework the event system to work via modes. Each mode represents a type of interaction the user has with the map data. Dragging, browsing, drawing a line, etc.
- move away from bucket based styling and into attribute styling. Hopefully this will let users have their own style for the different states and let users create as many states as they want.
- make private apis private
- standardize api verbs (add/remove, vs add/destroy) as much as we can.
- speed things up


lucaswoj:
Exciting! Ping me if you want a code review once this nears a shippable state.

src/events.js
```diff
@@ -0,0 +1,97 @@
+var ModeHandler = require('./modes/mode_handler');
+var browse = require('./modes/browse');
+var findTaragetAt = require('./lib/find_target_at');
```
kelvinabrokwa: `s/Taraget/Target/g`

kelvinabrokwa:
Any particular reason for moving away from ES6 syntax?

mcwhittemore:
@kelvinabrokwa classes make hiding private vars via closures pretty hard, but most of the other changes are just my default writing style coming through. In time I'd like to find which syntax is best for perf and go with that in the places it maters. Please feel free to suggests syntax changes you think would help.

mcwhittemore:
I've spent the day trying to outline the api this branch is working towards into the api.md doc. I'd love some input on this as I continue to work. It is a pretty big shift in how Draw works so making sure it lets you do things like `select a feature` without requiring too much of the developer is key.

https://github.com/mapbox/mapbox-gl-draw/blob/b97cd9fc3a1cb24aeff596690d75215bbbde8e38/API.md

Thanks!

averas:
Great progress @mcwhittemore! I will give it a more thorough review but one thing that immediately caught my attention is the somewhat ambiguous description of id's. It says in the description of the <code>add</code> method that it will return <code>feature.id</code>, but in the examples later on it's clear that the id is present on the geometry rather than the top level feature. Is there any particular reason you've decided to maintain id's on the geometries of the features rather than the features themselves btw? If the Draw API always maintains and exposes features regardless of what GeoJSON you provide it's perhaps more intuitive to maintain the id's on the features.

joedjc:
@mcwhittemore Will the new styling for points allow for sprites? I think from a user perspective, lets say you are editing a layer which is styled with a icon/sprite, you'd probably expect when you add a new point that it looks the same and therefore creates an icon/sprite rather than a circle.

API.md
```diff
@@ -25,41 +25,16 @@ drawing | boolean | The ability to draw and delete features - default: `true`
...
 ## API Methods
 
-####`.add(Object: GeoJSONFeature, [Object]: options) -> String`
+`mapboxgl.Draw()` returns an instance of the `Draw` class which has the following public API methods for getting and setting data:
 
-This method takes any valid GeoJSON and adds it to Draw. The object will be turned into a GeoJSON feature and will be assigned a unique `drawId` that can be used to identify it. This method return the new feature's `drawId`. If an id is provided with the feature that ID will be used.
+###`.add(Object: GeoJSONFeature) -> String`
 
-The second argument is an optional options object to add information to the geometry when creating the new element. Currently the only used option is `permanent`, which, if set to true, will cause the element to ignore click events, `Draw.select(:id:)` and `Draw.selectAll()`.
+This method takes any valid GeoJSON and adds it to Draw. The object will be turned into a GeoJSON feature and will be assigned a unique `id that can be used to identify it. This method return  `feature.id`. If an id is provided with the feature that ID will be used.
```
kelvinabrokwa: s/return/returns (tiny typo)

API.md
```diff
@@ -25,41 +25,16 @@ drawing | boolean | The ability to draw and delete features - default: `true`
...
 
 Draw does not enforce unique IDs to be passed to `.add`, but it does enforce unique ids inside of it. This means that if you provide an id for a feature that is not unqiue, Draw will override the exhisting feature with your new feature. You can think of this like PUT in http verbs.
```
kelvinabrokwa: s/exhisting/existing/

API.md
```diff
@@ -139,121 +113,178 @@ console.log(Draw.getAll());
...
+  .getAll();
+// => { type: 'FeatureCollection', features: [] }
+```
 
 ---
 
-####`.deselect(String: drawId) -> Draw`
+### `.trash() -> Draw`
+
+This envokes the current modes trash event. For the `default` mode this deletes all active features. For the `direct_select` mode this deletes the active vertecies. For the draw modes, these cancels the current process.
```
kelvinabrokwa: s/vertecies/vertices

API.md
```diff
@@ -139,121 +113,178 @@ console.log(Draw.getAll());
...
+This is fired every time a feature is deleted inside of `mapbox-gl-draw`. The payload is the GeoJSON feature just before it was deleted.
 
-This method initiates drawing the specified type. The types are defined in constants in draw under `Draw.types.*`. The argument should be one of
+### draw.active
 
-- `Draw.types.POINT`
-- `Draw.types.LINE`
-- `Draw.types.POLYGON`
-- `Draw.types.SQUARE`
+This is fired every time a feature is set to active or inactive. If a feature was active and is then set to active again, this is not fired. Same goes for inactive. This is only fired for features. Not for verticeies.
```
kelvinabrokwa: s/verticeies/vertices

API.md
```diff
@@ -139,121 +113,178 @@ console.log(Draw.getAll());
...
+## Styling Draw
 
-Draw fires off a number of events on draw and select actions. All of these events are name spaced `draw` inside of the mapboxgl event emitter.
+Draw is styled by the [Mapbox GL Style Spec](https://www.mapbox.com/mapbox-gl-style-spec/) with a preset list of properties.
 
-#### draw.set
+The `GL Style Spec` requires each layer to have a source. **DO NOT PROVIDE THIS** for styling draw.
 
-This is fired every time a feature is committed via escape or the double click. The payload is an object with the `mapbox-gl-draw` feature id and the GeoJSON representation of the feature.
+Draw moves features between sources for performence gains, because of this it is recommeneded that you **DO NOT** provide a source for a stlye despite the fact the `GL Style Spec` requires a source. **Draw will provide the source for you automaticlly**.
```
kelvinabrokwa: s/stlye/style

s/performence/performance

s/automaticlly/automatically

index.js
```diff
@@ -1,44 +1,31 @@
...
- * var mapboxgl = require('mapbox-gl');
- * var GLDraw = require('mapbox-gl-draw');
- * var map = new mapboxgl.Map({
- *   container: 'map',
- *   style: 'https://www.mapbox.com/mapbox-gl-styles/styles/outdoors-v7.json'
- * });
- * var Draw = GLDraw();
- * mapboxgl.addControl(Draw);
- */
-import Draw from './src/draw';
```
kelvinabrokwa: this docs could come in handy if this gets big enough that we need automatic doc generation

package.json
```diff
@@ -41,7 +41,8 @@
     "mapbox-gl-js-mock": "^0.15.0",
     "mock-browser": "^0.92.10",
     "tape": "^4.0.0",
-    "uglify-js": "^2.4.23"
+    "uglify-js": "^2.4.23",
+    "mapbox-gl": "https://github.com/mapbox/mapbox-gl-js/tarball/a957acdb16952dbea30bf8d000480e7888578772"
```
kelvinabrokwa: whats in this version specifically that we need?

src/api.js
```diff
@@ -1,199 +1,102 @@
-'use strict';
```
kelvinabrokwa: why not `'use strict'`?

mcwhittemore:
> one thing that immediately caught my attention is the somewhat ambiguous description of id's. It says in the description of the add method that it will return feature.id, but in the examples later on it's clear that the id is present on the geometry rather than the top level feature.

@averas thanks for point this out. I'll try to clear up ids later today and ping you for feedback when I think I have a good handle on it.

> To minimise confusion it's perhaps also best to either base the examples on features, or rename the variables to "geojson" etc., where you in fact are not passing in features.

@averas yea, I was feeling feature fatigue when I wrote this. GeoJSON is a good move. I think we're going to need to solve this in the code too.

> Will the new styling for points allow for sprites?

@joedjc Yep. I'm pretty sure this is currently possible as well. Draw doesn't ship with sprites though, so you'd need to make sure they are included at the map level.

This below style should put an icon on all Point features in the *future* version of gl-draw. In the current version you'd need to name the layer to overwrite the point layer which also means your LineString and Polygon handles would be this icon.

```json
{
    "id": "coffee-shops",
    "type": "symbol",
    "filter": ["all",
      ["==", "$type", "Point"],
      ["==", "meta", "feature"]],
    "layout": {
        "icon-image": "icon_name"
    },
    "interactive": true
}
```

package.json
mcwhittemore: `queryRenderFeatures` become sync and faster in this version. This lets us get features in the same event loop as the mouse event was triggered on so we can stopPropagation if we want. There were also race conditions between mouse events with featuresAt.

src/api.js
mcwhittemore: I might have just deleted it on accident. That said, I'd like to have babel do this for us so we don't have to remember.

https://babeljs.io/docs/plugins/transform-strict-mode/

index.js
mcwhittemore: @kelvinabrokwa agreed. Doc comments that aren't used get stale quickly though so I removed them rather than updating them.

drewbo:
@mcwhittemore very cool stuff. A few things I noticed:

- Is there a way to get all active features now? (formerly `getSelected()`) If so, this makes the `.trash` method more of convenience function

- Would prefer all events to be lower case, no underscore, to match mapbox-gl (e.g. `boxzoomend`)

- Would prefer modes to be either enums (`Draw.modes.DEFAULT`) or separate methods (`Draw.startDirectSelect(id)`, `Draw.startDrawing(Draw.types.LINE)`) although then I'm not positive what `draw.changemode` should emit.

- There's no mode which doesn't select a feature when clicked (formerly `permanent` as an option but I agree with the move to remove these properties from features). This could be accomplished by moving a feature out of the draw source but I can't decide if that should be handled by the library or developer

- `draw.active` is an ambiguous event name. Would prefer two events (`draw.activate` and `draw.deactivate` or something like that) or one event with a broader name `draw.changeInActiveness` (maybe a little verbose...)

mcwhittemore:
> Is there a way to get all active features now? (formerly getSelected()) If so, this makes the .trash method more of convenience function

I don't currently have a `getSelected()`. I think the getters should only return features drawn or provided by the user. Since you also select vertices and maybe edges (in the future) I think `getSelected` or even `getActive` is confusing.

Assuming we had a getter that returned the user provided feature the mode is currently focused on you still wouldn't be able to preform `trash` because `trash` is telling the current mode to run its deletion code.  For `direct_select` this would be the selected vertices, not the active feature. For the draw modes, its more of a cancel than a delete.

> Would prefer all events to be lower case, no underscore

:+1: 

> There's no mode which doesn't select a feature when clicked

Yea. I don't think GL draw should get into this game. If the apis for this aren't good for mapbox-gl-js we should fix it there or create a small module that makes this easy.

> draw.active is an ambiguous event name

Agree. There isn't an event for feature changes. Maybe we can solve these two problems with one event?

src/api.js
kelvinabrokwa: :+1:

src/setup.js
```diff
@@ -0,0 +1,99 @@
...
+            features: []
+          },
+          type: 'geojson'
+        });
+
+        for (let i = 0; i < theme.length; i++) {
+          let style = theme[i];
+          style.source = 'draw-selected';
+          // TODO: this should be on both sources...
+          // TODO: let users overwrite this...
```
kelvinabrokwa: @mcwhittemore are you going to hit this on this PR?

src/render.js
```diff
@@ -0,0 +1,33 @@
...
+    var featureBuckets = Object.keys(this.features).reduce((buckets, id) => {
+      let featureInternal = this.features[id].internal();
+      featureInternal.properties.mode = mode;
+      let modeFeatures = this.ctx.events.currentModeRender(featureInternal);
+
+      if (!Array.isArray(modeFeatures)) {
+        modeFeatures = [modeFeatures];
+      }
+
+      if (true) {
```
kelvinabrokwa: constant conditional?

src/options.js
```diff
@@ -0,0 +1,34 @@
...
+const hideControls = {
+  point: false,
+  line: false,
+  polygon: false,
+  trash: false
+};
+
+module.exports = function(options = {controls: {}}) {
+
+  if (options.displayControlsDefault === false) {
```
kelvinabrokwa:
```js
if (!options.displayControlsDefault)
```

(just being a bit picky)

src/modes/draw_polygon.js
```diff
@@ -0,0 +1,91 @@
...
+    ctx.events.changeMode('default');
+    ctx.store.delete(feature.id);
+  }
+
+  var pos = 0;
+
+  var onMouseMove = function(e) {
+    if(pos === 0) {
+      feature.updateCoordinate(`0.${0}`, e.lngLat.lng, e.lngLat.lat);
+      feature.updateCoordinate(`0.${1}`, e.lngLat.lng, e.lngLat.lat);
```
kelvinabrokwa: what's the point of this string formatting?

src/modes/draw_polygon.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+    else {
+      feature.updateCoordinate(`0.${pos}`, e.lngLat.lng, e.lngLat.lat);
+    }
+  }
+
+  var onClick = function(e) {
+    // did we click on the last point
+    // did we click on the first point
```
kelvinabrokwa: to do?

src/modes/draw_polygon.js
```diff
@@ -0,0 +1,91 @@
...
+    else {
+      ctx.events.changeMode('default');
+    }
+  }
+
+  return {
+    start: function() {
+      ctx.ui.setClass('mapbox-gl-draw_mouse-add');
+      this.on('mousemove', () => true, onMouseMove);
+      this.on('click', () => true, onClick);
```
kelvinabrokwa: what does this first callback function do?

src/api.js
```diff
@@ -1,199 +1,72 @@
...
-   * @param {String} id - the draw id of the geometry
-   */
-  get(id) {
-    var feature = this._store.get(id);
-    return feature && feature.toGeoJSON();
-  }
+var featureTypes = {
+  "Polygon": require('./feature_types/polygon'),
+  "LineString": require('./feature_types/line_string'),
+  "Point": require('./feature_types/point')
```
kelvinabrokwa: single quote

src/api.js
```diff
@@ -1,199 +1,72 @@
...
+}
```
kelvinabrokwa: semicolon

src/api.js
```diff
@@ -1,199 +1,72 @@
...
-    return this._store.getAllIds()
-      .map(id => this._store.get(id).toGeoJSON())
-      .reduce(function(featureCollection, feature) {
-        featureCollection.features.push(feature);
-        return featureCollection;
-      }, {
+var API = module.exports = function(ctx) {
+
+  return {
+    add: function (geojson, opts) {
```
kelvinabrokwa: the opts param is not used

src/api.js
```diff
@@ -1,199 +1,72 @@
...
+      var geojson = JSON.parse(JSON.stringify(geojson));
+      if (geojson.type === 'FeatureCollection') {
+        return geojson.features.map(feature => this.add(feature, options));
```
kelvinabrokwa: should options be opts?

src/setup.js
```diff
@@ -0,0 +1,99 @@
...
+module.exports = function(ctx) {
+
+  ctx.events = events(ctx);
+
+  ctx.map = null;
+  ctx.container = null;
+  ctx.store = null;
+  ui(ctx);
+
+  var buttons = {};
```
kelvinabrokwa: buttons is not used here

src/ui.js
```diff
@@ -0,0 +1,103 @@
...
+      if (buttons[id]) {
+        buttons[id].style.display = 'block';
+      }
+    },
+    setButtonActive: function(id) {
+      if (buttons[id] && id !== 'trash') {
+        buttons[id].classList.add('active');
+      }
+    },
+    setAllInactive: function(id) {
```
kelvinabrokwa: the `id` param is unused. What's the purpose?

src/lib/mode_handler.js
```diff
@@ -0,0 +1,82 @@
...
+    doubleclick: [],
+    mousemove: [],
+    mousedown: [],
+    mouseup: [],
+    keydown: [],
+    keyup: [],
+    trash: []
+  };
+
+  var lastClass = '';
```
kelvinabrokwa: what is this used for?

kelvinabrokwa:
@mcwhittemore this could really use a linting pass

src/setup.js
mcwhittemore: Yep

src/render.js
mcwhittemore: Currently I'm just working with one source - this PR is very much a WIP.

src/options.js
mcwhittemore: Yea, I'd much rather have it be `=== false`. It is only false if false is sent, 0, undefined and null don't set it.

src/modes/draw_polygon.js
mcwhittemore: The point here is to match what `queryRenderFeatures` returns. With VT the properties are either a number or a string. I assumed that was the same for GeoJSON sent to GL, but I didn't check this. I wonder if we could send an array.

src/modes/draw_polygon.js
mcwhittemore: If they return false the next function does not run. If they return true the next function does run and no other event handler is tested.

mcwhittemore:
Just a few bugs

- [ ] dragging often draws to drag the map like its an image...
- [ ] last mid-point is missing on polygons in direct_select mode
- [ ] trash can is not showing
- [ ] single click in direct_select mode should bring you back to select mode when no coords are selected
- [ ] double click zooms...
- [ ] shift select is default mode doesn't work
- [ ] selecting solid lines is hard
- [ ] select dotted lines is really hard
- [ ] key bindings don't work

mcwhittemore:
- [ ] get events working
- [ ] remove tests that cover dropped apis
- [ ] work on clustering 'invisible features'

bench/index.js
```diff
@@ -0,0 +1,139 @@
...
+
+    benchmark.on('pass', function(event) {
+        log('green', '<strong class="prose-big">' + event.message + '</strong>');
+        scrollToBottom();
+    });
+
+    benchmark.on('fail', function(event) {
+        log('red', '<strong class="prose-big">' + event.message + '</strong>');
+        scrollToBottom();
+    });
```
lucaswoj: What does it mean for a benchmark to "pass" or "fail"? Don't you want a metric that you can work to improve over time?

PR #259, opened 2016-03-16
Fix for issue when you drag a shape and instead of a shape map is dragged

kiRach:
Hi guys,

I've found a serious issue while was working with this plugin. Let me try to explain what I mean. 

So, I place a shape on a map and make it `interactive: true` to always have it in editable state. And when I try to move shape on a map (the cycle is: mouse down on the shape -> move mouse -> mouse up) sometimes it works ok, but sometimes I drag the map itself instead of dragging the shape. 

I started investigate and that's what I found: when we receive `mouseDrag` event - we check if there is `activeDrawId` here: https://github.com/mapbox/mapbox-gl-draw/blob/dca6a21050f01995c61c80d3eaacaabc4eebc863/src%2Fdraw_events.js#L125. And if there is - we stop propagating event to move the shape (and not move the map). But the issue is that we set this  `activeDrawId` in handler for `mouseDown` event in asynchronous callback for 'mapbox.featruresAt' method:  https://github.com/mapbox/mapbox-gl-draw/blob/dca6a21050f01995c61c80d3eaacaabc4eebc863/src%2Fdraw_events.js#L70

Since it asynchronous, what happens sometimes: we receive `mouseDown` event and then immediately receive `mouseDrag` event. And checking for `if (activeDrawId)` is false because `featuresAt` is asynchronous and wasn't finished yet. And then we ignore dragging event and as a result - user drags map, not shape.

The flow is like here:
-> received `mouseDown` event
-> call for `mapbox.featruresAt`
-> received mouseDrag event
-> check for `activeDrawId` - false - ignore `mouseDrag` event
-> asynchronous callback `mapbox.featruresAt` is called
-> set `activeDrawId`

In this PR I made a quick fix for this issue. May be it's not best way to resolve this issue, but it works for me.

### DIFF
--- a/src/draw_events.js
+++ b/src/draw_events.js
@@ -108,7 +108,9 @@ export default function(ctx) {
     },
     onMouseMove: function(e) {
       if (isMouseDown) {
-        featuresAtProgress && e.originalEvent.stopPropagation();
+        if (featuresAtProgress) {
+          e.originalEvent.stopPropagation();
+        }
         return api.onMouseDrag(e);
       }
       else if (newFeature) {
--- a/src/draw_events.js
+++ b/src/draw_events.js
@@ -16,6 +16,7 @@ export default function(ctx) {
   var newFeature = null;
   var activeVertex = null;
   var activeDrawId = null;
+  var featuresAtProgress = false;
 
   var cleanupNewFeatureIfNeeded = function() {
      if (newFeature.created) {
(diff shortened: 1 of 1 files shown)


mcwhittemore:
@kiRach - thanks for looking into this. I'm currently working on a pretty big internal refactor to help clean up our event system and make advanced drawing modes simpler to reason about. To make sure my changes don't add more bugs I'm trying to make sure all bug fixes are tested before merging. Can you write a test that confirms this works as expected?

Also, this is failing our lint settings. You can run `npm run lint` to see these errors before pushing.

kiRach:
@mcwhittemore I fixed lint warning. What about test - really have no idea how to test it. It depends on user interactions - mouse down, mouse drag events and also this issue is not 100% reproducible. If callback for `featuresAt` method was called before `onMouseDrag` event - then it works good. If `onMouseDrag` was before callback - then there is an issue.

mcwhittemore:
Yea... this would need an upstream fix in the mapbox-gl-js-mock lib as well. The featuresAt functionally in the mock should call a callback after x ms. With this in mind you would use `map.fire` to 'mouse events'. If you fire faster than the setTimeout you would under-fire the race condition. If you fired slower than the setTimeout you would over-fire the race condition.

If its something you want to do (for fun) that would be great. But I agree its a bit of an ask for this PR.

mcwhittemore:
@kiRach this is in 0.4.0 on npm. Thanks again!

PR #273, opened 2016-03-30
Only emit valid changes, don't emit empty changesets, don't update on add if nothing has changed

mcwhittemore:


mcwhittemore:
@scothis for the review

PR #280, opened 2016-04-01
Exports fix + position option implementation

nsamala:
I kept getting a blank object when I tried using `require('mapbox-gl-draw')` with Browserify. I changed the way the module exports itself and it works now.

I also added code to support multiple ui positions similar to `mapboxgl.Navigation`

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -73,7 +73,7 @@ location.hash.replace(/^#/,'').split('/').reduce(function(args, val, i, hash){
     position: 'top-left'
   }));
 
-  var Draw = window.Draw = mapboxgl.Draw();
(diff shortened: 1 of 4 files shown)


mcwhittemore:
@nsamala thanks for this. Can you get the tests passing for the options ui?

nsamala:
Sure thing! Looks like it was mainly some trailing spaces. Ran it locally so here's hoping :)

mcwhittemore:
@nsamala - one small bug. If I try put the navigation control in a different corner than the draw control the draw control doesn't show up.

You can test this by tinkering with the debug/index.html file.

nsamala:
Ah yeah I noticed that issue with the original control. I just pushed up a fix for that. It now works even in the absence of a navigation control

PR #283, opened 2016-04-07
Fix draw.changed event payload to return only changed features

lukeelmers:
I noticed that the `draw.changed` event would return all of the existing features in the payload. This update fixes the behavior so it will only return the feature that was changed.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -53,14 +53,13 @@ module.exports = function render() {
     var changed = [];
 
     this.featureIds.forEach((id) => {
-      let featureInternal = this.features[id].internal(mode);
+      let feature = this.features[id];
+      let featureInternal = feature.internal(mode);
       var coords = JSON.stringify(featureInternal.geometry.coordinates);
 
-      if (this.ctx.store.needsUpdate(featureInternal)) {
+      if (feature.isValid() && this.ctx.store.needsUpdate(feature.toGeoJSON())) {
         this.featureHistory[id] = coords;
-        if (this.features[id].isValid()) {
-          changed.push(this.features[id].toGeoJSON());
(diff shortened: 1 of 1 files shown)


src/render.js
```diff
@@ -53,14 +53,13 @@ module.exports = function render() {
     var changed = [];
 
     this.featureIds.forEach((id) => {
-      let featureInternal = this.features[id].internal(mode);
+      let feature = this.features[id];
+      let featureInternal = feature.internal(mode);
       var coords = JSON.stringify(featureInternal.geometry.coordinates);
 
-      if (this.ctx.store.needsUpdate(featureInternal)) {
+      if (feature.isValid() && this.ctx.store.needsUpdate(feature.toGeoJSON())) {
```
mcwhittemore: I'd like to use `featureInternal` here so we don't `JSON.parse(JSON.stringify())` more than needed.

mcwhittemore:
@lukeelmers I haven't been seeing this. Can you explain what was causing the problem? Also, can you put together a test for this so it doesn't regress.

averas:
I am seeing the same kind of behaviour. First off, "draw.changed" is fired more or less all the time. If I simply drag the map view around the event will fire. When adding a new point and just moving the mouse around the event will fire... and as this PR suggests, all features are provided with any of these events.

mcwhittemore:
@lukeelmers thanks for this. I just rebased your commit and merged it via #293. Sorry it took so long, I've been traveling for the last couple weeks.

lukeelmers:
@mcwhittemore No prob! Sorry to go MIA on this thread for a couple weeks -- glad I could help.

PR #295, opened 2016-04-26
update feature properties on Draw.add even if there is no change in coords

### LINKED ISSUE #287: Changing a feature's properties and updating via draw.add does not work
Currently draw does not seem to consider mutation of properties a change so the "needsUpdate" is not triggered when passing in a feature with an existing id with non-updated coordinates but with updated properties. 

If you want to update the properties of an existing feature in the store you have to delete it and add it again...

mcwhittemore:
Closes #287 

@averas for the review

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "express": "^4.13.4",
     "geojson-validation": "^0.1.6",
     "mapbox-gl": "^0.17.0",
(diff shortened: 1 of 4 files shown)


averas:
Looks good and tested locally!

mcwhittemore:
Thanks!

PR #301, opened 2016-05-05
Synthetic clicks

mcwhittemore:
Treats clicks as mousedown + mouseup events that do happen without the mouse moving too far or too much time happening between them.

Fixes

The logic is borrowed from [iD](https://github.com/openstreetmap/iD/blob/d130c517b0a98c470aef269176bda729532f8ff1/js/id/behavior/draw.js#L34-L56).

@samanpwbb for the review.

### DIFF
--- a/bench/lib/mouse_draw.js
+++ b/bench/lib/mouse_draw.js
@@ -7,20 +7,23 @@ module.exports = function(start, map) {
 
   var events = mouseEvents(map);
 
+  events.push('mousedown', {
+    x: start.x,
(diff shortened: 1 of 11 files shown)


samanpwbb:
Solves the main problem. I notice that the map will still take a small drag. Maybe if a point is going to be added, then prevent the drag from propagating to the map?

mcwhittemore:
@samanpwbb I've stopped all drag events that could turn out to be click events... I think this should do it.

Issue #310, opened 2016-05-16
Invalid LngLat object: (undefined, undefined) after direct_select Point feature


stepankuzmin:
Changing mode to `direct_select` with id of feature with `Point` geometry type causes `Uncaught Error: Invalid LngLat object: (undefined, undefined)`

![2016-05-16 14 01 39](https://cloud.githubusercontent.com/assets/533564/15285111/c94cb00e-1b6e-11e6-9c9d-a30082de4e87.png)

**mapbox-gl-draw v0.6.2**

mcwhittemore:
@stepankuzmin this is because `direct_select` doesn't handle Points. A different error message should be thrown that explains this though.

PR #311, opened 2016-05-17
Throw error when using direct_select with points

stepankuzmin:
Throw error when using `direct_select` with points as mentioned in mapbox/mapbox-gl-draw#310


mcwhittemore:
This has been published as 0.6.3. Thanks for the pr.

PR #315, opened 2016-05-27
Ensure keybindings option is enforced

lukeelmers:
It appears the keybindings option is not checked before event listeners are attached, so `keybindings: false` has no effect. I've added a check for it, and things are behaving as expected now.

Let me know if there's another way you'd prefer to approach this -- I'm happy to update this PR as needed.


mcwhittemore:
This looks great. Thanks.

mcwhittemore:
`0.6.5` has been released with this fix @lukeelmers

PR #321, opened 2016-06-02
Add draw.mode.simple_select.selected events

lukeelmers:
It appears that the `draw.mode.simple_select.selected.start` and `draw.mode.simple_select.selected.end` events [mentioned in the docs](https://github.com/mapbox/mapbox-gl-draw/blob/master/API.md#drawmodesimple_selectselectedstart) have not yet been implemented, as the existing `selected.start` and `selected.end` events are just used internally and not fired on `ctx.map`. This PR adds support for both events.

### DIFF
--- a/API.md
+++ b/API.md
@@ -232,11 +232,11 @@ This is fired when feature coordinates are changed. It is fired just after the c
 
 ### draw.mode.simple_select.selected.start
 
-This is fired every time a feature is selected in the default mode. The payload is an array of feature ids being selected. This is **NOT** fired when the mode starts as this information is in the `draw.modechange` event.
+This is fired every time a feature is selected in the default mode. The payload is an object containing the `featureIds` being selected. This is **NOT** fired when the mode starts as this information is in the `draw.modechange` event.
 
 ### draw.mode.simple_select.selected.end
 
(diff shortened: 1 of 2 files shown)


mcwhittemore:
@lukeelmers sorry for the slow response. `this.fire` in modes is used to make sure that all events fired by a mode are prefixed with `draw.mode.MODE_NAME` so what you have added is emitting the same event twice.

Are you not seeing these events?

mcwhittemore:
Ah. I see the confusion now. The docs refer to this as `draw.mode.simple_select` and the helper wraps it as `draw.simple_select`.

https://github.com/mapbox/mapbox-gl-draw/blob/7f3632ac5e8d27d5c74ca729f347c9b830c90335/src/lib/mode_handler.js#L32

If you want to update the docs that would be great. If not, I'll do it in the next few days. Sorry for the confusion.

lukeelmers:
That makes much more sense. I totally missed where it was wrapped by the helper -- thanks for pointing this out!

mcwhittemore:
@lukeelmers not sure if you've seen, but we are working on making these events a lot better here https://github.com/mapbox/mapbox-gl-draw/issues/397

Issue #296, opened 2016-04-26
Performance of mapbox-gl-js degrades when draw is enabled


averas:
I noticed this when reviewing/testing #295...

1. Go to http://firefly.ibk.se/~racer/drawtest/
2. Zoom the map in/out and get a feel for the current performance (which is mostly related to screen size)
3. Enable draw and do the same

My impression is that overall performance degrades quite significantly, even though no features are added to the draw sources at all. Must be some hook in draw that's stealing CPU cycles.

averas:
I can't see anything obvious related to draw in the profiler.. but the CPU impact is definitely higher.. and the performance difference is evident, again, without any data on the sources at all.

Any clues here @mourner @lucaswoj ?

mcwhittemore:
Draw uses the zoom level to render features differently. This means that on zoom we're rerendering which, even with an empty dataset, would call setData on at least one source.

mourner:
@mcwhittemore can you expand on that? What differences exactly require setData on every zoom frame?

mcwhittemore:
@mourner Draw turns LineStrings and Polygons that are too small to be seen at the current zoom level to points which are styled as circle and can be seen at every zoom level. To do this we need to evaluate the size of a feature at the current zoom level. This is done on zoom end. https://github.com/mapbox/mapbox-gl-draw/blob/master/src/events.js#L144

mourner:
@mcwhittemore empty setData on zoomend doesn't seem like it should interfere with performance in any way. Anything else that GL draw hooks up?

mcwhittemore:
All the event handlers are here.

https://github.com/mapbox/mapbox-gl-draw/blob/master/src/events.js#L131-L145

There are a bunch of them.

mcwhittemore:
Oh, most of the mouse handlers are doing `queryRenderedFeatures` requests. We can probs cache this.

averas:
.. also <code>zoomend</code> would only be called on the "end of a zoom" I hope. :-) What I currently experience is that performance it hit _during_ zooming as well.

averas:
So, it's the <code>queryRenderedFeatures</code> fired as a result of <code>mousemove</code> I guess?

mcwhittemore:
@averas that seems like a good possibility.

averas:
[...]

PR #331, opened 2016-06-08
separate layers with hot and cold sources, refs #296

samanpwbb:
cc: @mourner @mcwhittemore @lucaswoj 

I also removed the fallback layer id assignment. Ids for style layers should be required.

I'll add tests if this makes sense to everyone.


src/options.js
```diff
@@ -24,6 +22,15 @@ const hideControls = {
   trash: false
 };
 
+function addSources(styles, temp) {
```
mcwhittemore: Lets call `temp` something like `sourceBucket`

samanpwbb: 🔥 ❄️

mcwhittemore:
This looks great @samanpwbb

samanpwbb:
Going to write some tests then merge 👍

samanpwbb:
Added tests, merging when green.

Issue #335, opened 2016-06-08
every single mouse move sets gl container class


samanpwbb:
This seems bad:

![screen shot 2016-06-08 at 5 06 55 pm](https://cloud.githubusercontent.com/assets/108094/15910937/565948ca-2d9b-11e6-8f8c-8d7f308383de.png)

Coming from https://github.com/mapbox/mapbox-gl-draw/blob/99e6ac6195f2e9ca5f3a8c6c42137db0fc5afebc/src/ui.js#L42-L45

mcwhittemore:
@samanpwbb is this updating the class every time or just looking to see if it needs to be updated?

samanpwbb:
Appears to be updating constantly.

PR #337, opened 2016-06-09
simplify how cursor classes get set, reducing noise while drawing, cl…

samanpwbb:
…oses #335

cc: @mcwhittemore


mcwhittemore:
@samanpwbb every time I click on the map while drawing mouse goes from `mouse-draw` to `mouse-add`.

PR #344, opened 2016-06-10
add box select option, closes #339

### LINKED ISSUE #339: if gl-draw is enabled, shift+drag should perform a bounding box multi-select
Most drawing tools include a way to select by creating a bounding box, and selecting all features inside the box via click and drag. 

By default, click + drag in mapbox gl js performs a bounding box zoom and that behavior is carried over by gl draw. 

On one hand, overriding default behavior is bad. On the other hand, we're already overriding some zoom controls (https://github.com/mapbox/mapbox-gl-draw/commit/5b9731072688b2770f29b75d3a05e9ab9b97b906), and bounding box select would be more useful than bounding box zoom while draw is enabled.

I can see some cases where bounding box zoom would be more intuitive, like if you're working on a tool where user toggles draw on and off a lot (and therefore toggles the way shift+drag works a lot), but for a more pure drawing tool, I have a hunch that bounding box select is expected behavior.

samanpwbb:
@mcwhittemore for review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "express": "^4.13.4",
     "geojson-validation": "^0.1.6",
     "mapbox-gl": "^0.19.0",
-    "mapbox-gl-js-mock": "^0.20.0",
+    "mapbox-gl-js-mock": "^0.21.0",
     "mock-browser": "^0.92.10",
     "tape": "^4.0.0",
     "uglify-js": "^2.4.23",
--- a/src/events.js
+++ b/src/events.js
@@ -50,7 +50,6 @@ module.exports = function(ctx) {
       time: new Date().getTime(),
       point: event.point
     };
-
     var target = getFeatureAtAndSetCursors(event, ctx);
     event.featureTarget = target;
     currentMode.mousedown(event);
--- a/src/events.js
+++ b/src/events.js
@@ -69,7 +69,6 @@ module.exports = function(ctx) {
     else {
       currentMode.mouseup(event);
     }
-
   };
 
   events.trash = function() {
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -46,6 +46,7 @@ var ModeHandler = function(mode, DrawContext) {
         break;
       }
     }
+    DrawContext.ui.fireClassUpdate();
   };
 
   mode.start.call(ctx);
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -3,7 +3,6 @@ var ModeHandler = function(mode, DrawContext) {
   var handlers = {
     drag: [],
     click: [],
-    doubleclick: [],
     mousemove: [],
     mousedown: [],
     mouseup: [],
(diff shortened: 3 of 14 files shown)


src/setup.js
```diff
@@ -32,6 +32,14 @@ module.exports = function(ctx) {
 
       ctx.ui.addButtons();
 
+      if (ctx.options.boxSelect) {
+        map.boxZoom.disable();
+        // Need to toggle dragPan on and off or else first
+        // dragPan disable attempt in simple_select doesn't work
+        map.dragPan.disable();
+        map.dragPan.enable();
+      }
```
samanpwbb: This ☝️ is weird. If I don't include the dragPan disable/enable toggle, map.dragPan.disable() will not work the first time it's called in simple_select. It will work fine all subsequent times.

samanpwbb:
@mcwhittemore for final review

package.json
```diff
@@ -45,7 +45,7 @@
     "express": "^4.13.4",
     "geojson-validation": "^0.1.6",
     "mapbox-gl": "^0.19.0",
-    "mapbox-gl-js-mock": "^0.20.0",
+    "mapbox-gl-js-mock": "^0.21.0",
```
mcwhittemore: mapbox-gl-js-mock should match the version of mapbox-gl-js...

samanpwbb:
Found a bug when selecting things with click + drag followed by shift + click:

screen shot 2016-06-10 at 1 55 11 pm

src/events.js
```diff
@@ -96,12 +94,14 @@ module.exports = function(ctx) {
...
       ctx.api.changeMode('draw_polygon');
     }
+    ctx.ui.fireClassUpdate();
   };
 
   events.keyup = function(event) {
     if (isKeyModeValid(event.keyCode)) {
       currentMode.keyup(event);
     }
+    ctx.ui.fireClassUpdate();
```
mcwhittemore: You should be able to drop this as this will be caught by the mode_handler delegator.

mcwhittemore: Ah... so the problem is that we change class without the handlers firing...

package.json
samanpwbb: I added some missing methods and incremented the version.

mapbox-gl-js-mock is not anywhere close to complete, and it's not documented anywhere that the two projects are supposed to have the same version number.

src/events.js
samanpwbb: oops, i forgot to remove this!

src/lib/mode_handler.js
```diff
@@ -47,6 +46,7 @@ var ModeHandler = function(mode, DrawContext) {
         break;
       }
     }
+    DrawContext.ui.fireClassUpdate();
```
mcwhittemore: Since we need to do this for all events in events.js we can drop this.

src/events.js
mcwhittemore: ok, I lied in the first comment. We need to do this for each event type in events.js and we should do it anywhere else in Draw.

src/lib/mode_handler.js
samanpwbb: I think it's better to only have it here.

package.json
mcwhittemore: yea, I just looked at the README and realized that I didn't push the draft that talked about versioning. Sorry.

PR #356, opened 2016-06-13
prevent adding extra nodes to polygons, fixes #345

### LINKED ISSUE #345: Sometimes final point in polygon ends up with many vertices
I clicked 4 times to create a polygon. Then, I tried to drag the top corner of the polygon. I was able to pull a 5 new vertices out of what seemed like one vertex:

screen shot 2016-06-10 at 10 16 00 am

I do not have a consistent way to reproduce this bug.

samanpwbb:
@mcwhittemore for review

### DIFF
--- a/src/feature_types/polygon.js
+++ b/src/feature_types/polygon.js
@@ -1,5 +1,11 @@
 var Feature = require('./feature');
 
+function firstMatchesLast(coords) {
+  return coords.every(coord => coord[0].every((el, i) => {
+    return el === coord[coord.length - 1][i];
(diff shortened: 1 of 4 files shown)


mcwhittemore:
@samanpwbb what if the geojson has two positions that are identical legitimately? I don't see anything in the spec that says this can't happen.

samanpwbb:
@mcwhittemore I think we should remove the error throws, use comments to clarify the intent of these two functions, and write tests against the `buildFeatureCoords` function.

Issue #314, opened 2016-05-26
Better support for finalising drawn shapes


aidanlister:
My use case is the user creates a bounding box and I highlight all the markers within that bounding box.

The Google Maps API has a draw.end style event when the user double clicks, allowing me to change the color of the bounded area and move to the next step. This works quite well.

I haven't been able to do this as elegantly with draw.changed ... it fires in realtime as the user changes their shape so I either need to debounce it or try and add a "finished" button somewhere into the process.

What's the thinking on how this could be better supported?

mcwhittemore:
@aidanlister - just to make sure we're on the same page here. In your use case the user is drawing a polygon and you are using this polygon to highlight a bunch of points. Is that correct?

`draw.changed` gets emitted once per frame if something has changed. I don't think slowing this down inside of Draw is a good idea as it limits the speed someone could respond to changes. Playing around with your use case a bit it seems like debouncing is a great option. Why don't you want to do this?

Assuming there is a good reason debouncing doesn't work, there are a few other events that draw emits that help us guess at when someone is done editing.  `draw.modechanged` fires when the behavior of Draw is changed. When going to `simple_select` from any other mode we can assume drawing has stopped. That said, editing can happen in both `simple_select` and `direct_select`. If unselecting the feature is a good enough measurement of the end of drawing than we can use the `draw.mode.simple_select.selected.end` event along with `draw.modechanged` to know when drawing has ended. This has problems though, ie: I don't want to unselect the feature but I'm done drawing. My guess is that some sort of debounce + the above solution could get you 90% of the way there.

This is a great idea and is a problem worth solving but the `when is the user done drawing problem` is not the aim of Draw. This kind of interaction detection is very complex. If you want to create module that does this and need Draw to emit some details it is not emitting currently please feel free to ask (or open a PR with your additions). Also, if you make a module that does this, please link drop it here so that going forward users don't feel the need to reimplement. In the future (I'd guess) there will be a handful of utility tools like this that we will want to promote.

Good luck!

aidanlister:
@mcwhittemore Yes that's right, here's a screenshot of the process in google maps: http://share.aidanlister.com/gHrz

I also sent you a link to the sandbox if you wanted to play with it, no worries if you don't have time.

I see what you're saying though ... the nature of the polygon tool in mapbox-gl-draw is quite different as it's immediately a valid shape from the first click. The google maps polygon tool is only valid once you've double clicked to complete the shape.

[...]

PR #358, opened 2016-06-13
Modified event

ghoshkaj:
Relates to https://github.com/mapbox/mapbox-gl-draw/issues/314. What I'm doing here is just storing the most recent set of features after a draw.changed event, and broadcasting that most recent set of features on a mouseUp.  So this relays a message on every click while drawing a new feature, and for updating vertices on existing features.

I still need to write tests and make the existing broken ones pass.
cc/ @mcwhittemore

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -65,7 +65,8 @@ module.exports = function(ctx) {
     var target = getFeatureAtAndSetCursors(event, ctx);
     event.featureTarget = target;
 
-    ctx.map.fire('draw.modifiedFeature', {features: featuresChanged});
+    if (featuresChanged.length > 0) ctx.map.fire('draw.modified', {features: featuresChanged});
+    featuresChanged = [];
 
     if (isClick(mouseDownInfo, {
       point: event.point,
--- a/src/events.js
+++ b/src/events.js
@@ -19,6 +19,7 @@ module.exports = function(ctx) {
   var events = {};
   var currentModeName = 'simple_select';
   var currentMode = ModeHandler(modes.simple_select(ctx), ctx);
+  var featuresChanged = [];
 
   events.drag = function(event) {
     if (isClick(mouseDownInfo, {
@@ -55,10 +56,16 @@ module.exports = function(ctx) {
     currentMode.mousedown(event);
   };
 
+  events.drawChanged = function(event) {
+    featuresChanged = event.features;
+  }
+
   events.mouseup = function(event) {
     mouseDownInfo.isDown = false;
     var target = getFeatureAtAndSetCursors(event, ctx);
     event.featureTarget = target;
+
+    ctx.map.fire('draw.modifiedFeature', {features: featuresChanged});
 
     if (isClick(mouseDownInfo, {
       point: event.point,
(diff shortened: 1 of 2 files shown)


mcwhittemore:
looking

src/events.js
```diff
@@ -55,11 +56,17 @@ module.exports = function(ctx) {
...
+  events.drawChanged = function(event) {
+    featuresChanged = event.features;
+  }
+
   events.mouseup = function(event) {
     mouseDownInfo.isDown = false;
     var target = getFeatureAtAndSetCursors(event, ctx);
     event.featureTarget = target;
 
+    ctx.map.fire('draw.modifiedFeature', {features: featuresChanged});
```
mcwhittemore: Lets call this `draw.modified` to match `draw.changed`.

mcwhittemore: if `featuresChanged.length === 0` lets not fire this event.

mcwhittemore: in https://github.com/mapbox/mapbox-gl-draw/pull/358/files#diff-4607e46b15da8ae44c4b47730c6d5963R48 we need to reset featuresChanged to [] or else we keep emitting the same features as changed.

debug/index.html
```diff
@@ -129,6 +129,8 @@
   req.send();
 }
 
+loadData('https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_50m_geography_regions_polys.geojson');
```
mcwhittemore: lets drop this from the debug

mcwhittemore:
@ghoshkaj this is looking good. Playing around with it a bit it looks like `draw.modified` will emit about 1 event for every 11 that `draw.changed` emits when drawing/dragging. I think this should be a big win for people.

src/events.js
ghoshkaj: Hey @mcwhittemore, so if I clear featureChanged on mousedown, it sends an empty array in mouseUp, as there is a mouseDown that comes immediately before every mouseUp. We don't have to worry about clearing it actually, because the draw.changed triggers on every single movement you make, and lines 59 to 62 ensures that featuresChanged is always storing the most up-to-date change coming from the users draw interaction.

ghoshkaj: We could send it fire the draw.modified change in  mouseDown instead, and then creal featureChanged in mouseUp. I'll experiment with that and see how that feels as an interaction.

mcwhittemore: @ghoshkaj when I was playing with this I found a bug where `draw.modfied` was firing (with features) even though `draw.changed` had not fired. We could also clear up `featuresChanged` just after firing. This might be the best idea.

ghoshkaj: @mcwhittemore good thinking, that is the quickest cleanest way. I'll do that! 😄

ghoshkaj:
@mcwhittemore addressed the changes in the review!

mcwhittemore:
@ghoshkaj this looks great. Can you update [the docs](https://github.com/mapbox/mapbox-gl-draw/blob/master/API.md#events) and than merge.

PR #361, opened 2016-06-14
Deselect deleted features; closes #357

### LINKED ISSUE #357: .delete() should clear selection
Seems like `draw.delete()` should clear the selection every time it's used. That way `draw.trash()` just needs to use `draw.delete()` and not manually deselect features.

Trouble is, the store code, of which `delete` is a part, does not currently have access to the selected features object in `simple_select.js`. So *I propose we stash the object of selected items on the store, instead of leaving it as a variable only available in `simple_select.js`.* 

Does that seem acceptable? I can put together a PR.

cc/ @mcwhittemore @samanpwbb

davidtheclark:
Ref #357.

This involved some refactoring to keep the selection in the store, accessible outside of simple_select. I performed some other other refactoring to clarify the API of the store and improve testability.

I tried a messing around on the debug map and everything seemed to me to work fine. Please check that I did not miss bugs or did not check everything.

cc/ @samanpwbb @mcwhittemore

### DIFF
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -54,8 +54,8 @@ var ModeHandler = function(mode, DrawContext) {
   return {
     render: mode.render || function(geojson) {return geojson; },
     stop: function() {
-      DrawContext.store.clearSelected();
       if (mode.stop) mode.stop();
+      DrawContext.store.clearSelected();
     },
     drag: function(event) {
       delegate('drag', event);
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -53,7 +53,10 @@ var ModeHandler = function(mode, DrawContext) {
 
   return {
     render: mode.render || function(geojson) {return geojson; },
-    stop: mode.stop || function() {},
+    stop: function() {
+      DrawContext.store.clearSelected();
+      if (mode.stop) mode.stop();
+    },
     drag: function(event) {
       delegate('drag', event);
     },
--- a/src/store.js
+++ b/src/store.js
@@ -1,61 +1,177 @@
 var {throttle} = require('./lib/util');
+var SimpleSet = require('./lib/simple_set');
 var render = require('./render');
 
 var Store = module.exports = function(ctx) {
+  this._features = {};
+  this._featureIds = new SimpleSet();
+  this._selectedFeatureIds = new SimpleSet();
+  this._changedIds = new SimpleSet();
   this.ctx = ctx;
-  this.features = {};
-  this.featureIds = [];
   this.sources = {
     hot: [],
     cold: []
   };
   this.render = throttle(render, 16, this);
-
   this.isDirty = false;
-  this.changedIds = [];
 };
 
+/**
+ * Sets the store's state to dirty.
+ * @return {Store} this
+ */
 Store.prototype.setDirty = function() {
   this.isDirty = true;
+  return this;
 };
 
-Store.prototype.featureChanged = function(id) {
-  if (this.changedIds.indexOf(id) === -1) {
-    this.changedIds.push(id);
-  }
+/**
+ * Sets a feature's state to changed.
+ * @param {string} featureId
+ * @return {Store} this
+ */
+Store.prototype.featureChanged = function(featureId) {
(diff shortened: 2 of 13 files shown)


src/store.js
```diff
@@ -1,61 +1,162 @@
...
   var deleted = [];
-  ids.forEach((id) => {
-    var idx = this.featureIds.indexOf(id);
-    if (idx !== -1) {
-      var feature = this.get(id);
-      deleted.push(feature.toGeoJSON());
-      delete this.features[id];
-      this.featureIds.splice(idx, 1);
-    }
+  [].concat(featureIds).forEach((id) => {
```
mcwhittemore: Why not just `forEach` over `featureIds`?

davidtheclark: Two reasons: So featureIds doesn't have to be an array, can just be a single value (which was preexisting API), and so the array is copied — because if i kept the original sometimes a screwy thing happened that I assume had to do with the array being mutated elsewhere.

mcwhittemore:
@davidtheclark this looks amazing! Thanks!

- [ ] in `simple_select` finish we should call `ctx.store.clearSelected`.
- [ ] in `direct_select` init we should select the selected feature and on finish we should call call `ctx.store.clearSelected`
- [ ] in `draw_*` we should select and clearSelected as the mode starts and finishes.

davidtheclark:
@mcwhittemore Can you describe the behaviors that I should see after making these changes correclty?

mcwhittemore:
getSelected() in the store should return the currently active feature for each mode.

davidtheclark:
Existing `getSelectedIds()` returns just the ids; so you'd like a `getSelected()` method that returns the selected feature objects?

mcwhittemore:
@davidtheclark sorry, misspoke, `getSelectedIds()` will do

davidtheclark:
@mcwhittemore Ok. Sorry but still not clear on how the interactions should change if I successfully fix the checkbox issues you mention above. How can I see what's wrong, and how then know I've fixed it?

mcwhittemore:
@davidtheclark this PR introduces the idea of selected features into the store. I think this is great as it will let us expose a way for end users to get what features are currently selected. 

Each mode defines what is selected so we need to make sure this new concept is applied to all modes. So far this PR applies it to simple_select.

For tests `Draw.changeMode` should be used to switch to a new mode. [The docs cover](https://github.com/mapbox/mapbox-gl-draw/blob/master/API.md#changemodestring-mode-any-options---draw) how to use this function to set what is active and not active.

davidtheclark:
@mcwhittemore The way that I ensured the selection was cleared whenever a mode changed was to add it to https://github.com/mapbox/mapbox-gl-draw/pull/361/files#diff-2650e3200d76013773ddf2af657f5a3dR57 instead of relying on each mode to clear the selection itself. I also set the selection in each mode's `start` function.

I wasn't sure if there were places in the existing unit tests that you thought I could add something ...

src/lib/mode_handler.js
```diff
@@ -53,7 +53,10 @@ var ModeHandler = function(mode, DrawContext) {
 
   return {
     render: mode.render || function(geojson) {return geojson; },
-    stop: mode.stop || function() {},
+    stop: function() {
+      DrawContext.store.clearSelected();
```
mcwhittemore: Lets do this after `mode.stop`, just in case a mode wants to do something with the selected features on end.

mcwhittemore:
@davidtheclark we don't have mode tests yet... we do need these though. But that is another PR.

src/lib/mode_handler.js
mcwhittemore: I just made this change.

mcwhittemore:
@davidtheclark I think this is ready to merge unless you had something else you wanted to do on it.

davidtheclark:
Thanks for making the change. Let's merge when it green-lights.

davidtheclark:
@mcwhittemore Should I add an entry to the changelog? Or do you like to do that all at once before release?

mcwhittemore:
@davidtheclark I like to have each merge release and add to change log. That way its easier to remember what changed.

PR #366, opened 2016-06-15
Emit new points in draw.modified and don't emit deleted features

### LINKED ISSUE #363: draw.modfied doesn't emit a point when it is created

### LINKED ISSUE #364: `draw.modified` shouldn't emit deleted features

mcwhittemore:
@davidtheclark for the review

Closes #364
Closes #363

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -20,17 +20,18 @@ module.exports = function(ctx) {
   var events = {};
   var currentModeName = 'simple_select';
   var currentMode = ModeHandler(modes.simple_select(ctx), ctx);
-  var changedFeatureIds = new SimpleSet();
+  var recentlyUpdatedFeatureIds = new SimpleSet();
 
   const emitModifiedFeatures = () => {
-    ctx.store.getChangedIds().forEach(id => changedFeatureIds.add(id));
+    ctx.store.getChangedIds().forEach(id => recentlyUpdatedFeatureIds.add(id));
 
-    let features = changedFeatureIds.values().map(id => ctx.store.get(id))
+    let features = recentlyUpdatedFeatureIds.values().map(id => ctx.store.get(id))
       .filter(f => f !== undefined)
+      .filter(f => f.isValid())
       .map(f => f.toGeoJSON());
 
     if (features.length > 0) ctx.map.fire('draw.modified', {features: features});
-    changedFeatureIds.clear();
+    recentlyUpdatedFeatureIds.clear();
   };
 
   events.drag = function(event) {
@@ -58,7 +59,7 @@ module.exports = function(ctx) {
   };
 
   events.drawChanged = function(event) {
-    event.features.forEach(f => changedFeatureIds.add(f.id));
+    event.features.forEach(f => recentlyUpdatedFeatureIds.add(f.id));
   };
 
   events.mousedown = function(event) {
--- a/src/events.js
+++ b/src/events.js
@@ -30,7 +30,9 @@ module.exports = function(ctx) {
       .filter(f => f.isValid())
       .map(f => f.toGeoJSON());
 
-    if (features.length > 0) ctx.map.fire('draw.modified', {features: features});
+    if (features.length > 0) {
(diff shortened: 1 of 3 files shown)


mcwhittemore:
I've been trying to write tests for this, but it looks like tests for event handlers are pretty sucky.

src/events.js
```diff
@@ -46,7 +58,7 @@ module.exports = function(ctx) {
   };
 
   events.drawChanged = function(event) {
-    featuresChanged = event.features;
+    event.features.forEach(f => changedFeatureIds.add(f.id));
```
davidtheclark: Why do we need another set for `changedFeatureIds` instead of just using the store's set directly? Seems odd to me that there are two sets with similar names but they are not in sync.

mcwhittemore: @davidtheclark The one is the store is cleared much more frequently than the one in events.js. Maybe we need a better name?

`changedFeatureIds` in events is all changed feature ids since the last time a modified event was triggered.

davidtheclark: As noted in https://github.com/mapbox/mapbox-gl-draw/issues/367, I think the "changed" vs "modified" situation is very confusing ... But certainly a more precise variable might help here, like `changedSinceLastModified` (or better word than "modified" I'd hope?)?

mcwhittemore:
I think this might emit all added features later on.

This event is really hard...

test/events.test.js
```diff
@@ -2,59 +2,177 @@
...
-    if(t.notDone()) {
-      done = true;
-      t.fail(msg || 'Timed out');
-      t.end();
-    }
-  })
+  t.equal('mousedown', map.fire.getCall(0).args[0]);
+  t.equal('mouseup', map.fire.getCall(1).args[0]);
+  t.equal('draw.modechange', map.fire.getCall(2).args[0]);
+  t.equal('draw.modified', map.fire.getCall(3).args[0]);
```
davidtheclark: Might want to add a test verifying the number of calls was only 4: `map.fire.callCount`

test/events.test.js
```diff
@@ -2,59 +2,177 @@
...
+      },
+      point: {x: 0, y:5},
+      lngLat: {lng: 0, lat: 5}
+  });
 
-  return t;
-}
+  var events = [
+    'mousedown',
+    'mouseup',
```
davidtheclark: `draw.modified` deliberately does not fire after the first vertex is created, right?

test/events.test.js
```diff
@@ -2,59 +2,177 @@
...
+  });
+
+  var events = [
+    'mousedown',
+    'mouseup',
+    'mousedown',
+    'mouseup',
+    'mousedown',
+    'mouseup',
+    'draw.modified',
```
davidtheclark: Again I'm just curious why it only fires twice.

test/events.test.js
```diff
@@ -2,59 +2,177 @@
...
+    var event = map.fire.getCall(0).args;
+    t.equal('draw.deleted', event[0]);
+    t.equal(1, event[1].featureIds.length);
+    // this event is weird, need to fix
+    //t.equal(id, event[1].featureIds[0]);
+
+    let numCalls = map.fire.callCount;
+    for(var i=0; i<numCalls; i++) {
+      var args = map.fire.getCall(i).args;
+      t.notEqual('draw.modified', args[0]);
```
davidtheclark: What is this check confirming?

davidtheclark:
Great! I had a couple of questions.

test/events.test.js
mcwhittemore: that `draw.modified` was not fired after deleting.

test/events.test.js
mcwhittemore: This is because it shouldn't fire when the feature isn't valid geojson.

test/events.test.js
mcwhittemore: It doesn't fire if the feature is not valid geojson

test/events.test.js
mcwhittemore: done

davidtheclark:
Thanks for the clarification! Looks good to me.

Issue #367, opened 2016-06-15
Standardize verb form of events


davidtheclark:
Right now we have -ed forms ("deleted") and plain forms ("modechange"). In the mapbox-gl-js API it seems that they are all plain forms ("rotate", "move", "mousestart", etc.). I think we should probably make all the draw events plain form verbs as well.

- [ ] Change `deleted` event to `delete`.
- [ ] Change `changed` event to `change`.
- [ ] Change `modified` event to `modify`. *I'll also add that the ambiguity between `changed` and `modified` is pretty confusing. There should be more distinction here, and if we're changing the form we may as well change the word to be more precise.*
- [ ] Change `selected` to `select`. This could be done as part of #365.

mcwhittemore:
I'm good with matching up with mapbox-gl-js.

The `change` vs `modify` events are really hard to distinguish in a short name because the only difference between them is how often they are fired. TBH, if we can get `draw.modify` to work well we could look at replacing `draw.change` with it.

Another name option is `draw.update`.

davidtheclark:
I think the shortness of the name is less important than having a clear semantic distinction between the events. Change and modify are essentially synonyms.

davidtheclark:
I don't see `draw.update` anywhere ... could you point to it in the code?

mcwhittemore:
@davidtheclark `draw.update` was a suggestion to replace `draw.modify`

davidtheclark:
Ah! Sorry :) Sure that's more clear. I guess I don't know what the actual distinction between events is yet, so not sure I'm the best judge.

davidtheclark:
Done in #407.

Issue #365, opened 2016-06-15
Better selection support


mcwhittemore:
Currently an end user needs to listen to the `draw.modechange`, `draw.simple_select.selected.start` and `draw.simple_select.selected.end` events to know what is selected. This came out of removing the general concept of 'selected' internally. With #361 Draw will knows what is selected at the store level rather than at that mode level again, meaning we can provide ways for users to get information about what is selected.

At first lets stay away from unselected features as this gets into forcing renders and switching modes.

1. `Draw.getSelectedIds` should return all selected feature ids of valid features.
1. `draw.selected` should fire when a feature is selected and is valid or when a selected feature becomes valid.
1. `draw.unselected` should fire when a valid feature is unselected.

For the events we want to make sure we don't trash `draw.unselected` and `draw.selected` with the same ids. To do this, I think we want to the events to fire with the render cycle rather than when their state has changed.

/cc @davidtheclark

davidtheclark:
> draw.selected should fire when a feature is selected and is valid or when a selected feature becomes valid.

Does this mean that you can select invalid features, with no event, and then if they become valid while they are selected an event is fired?

mcwhittemore:
@davidtheclark that comment was about how selection might work. I don't think Draw should emit invalid geojson, so its just a comment about that.

davidtheclark:
Closed by #368.

PR #368, opened 2016-06-16
New events

davidtheclark:
This first commit adds `draw.select` and `draw.deselect` events so addresses #365. They are unit tested using Sinon spies on the `map.fire` function. Also fixed some things in Store based on the tests that I wrote.

Do we want to to eliminate the `draw.mode.simple_select.selected.start` events, replacing with these new events? Or keep the mode-specific events also?

I could use this PR to remove those and follow through with an event cleanup re: #367.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -1,72 +1,85 @@
 module.exports = function render() {
-  var mapExists = this.ctx.map && this.ctx.map.getSource('mapbox-gl-draw-hot') !== undefined;
-  if (mapExists) {
-    var mode = this.ctx.events.currentModeName();
-    this.ctx.ui.setClass({
-      mode: mode
-    });
+  const store = this;
+  var mapExists = store.ctx.map && store.ctx.map.getSource('mapbox-gl-draw-hot') !== undefined;
+  if (!mapExists) return cleanup();
 
-    var newHotIds = [];
-    var newColdIds = [];
+  var mode = store.ctx.events.currentModeName();
+  store.ctx.ui.setClass({
+    mode: mode
+  });
 
-    if (this.isDirty) {
-      newColdIds = this.getAllIds();
+  var newHotIds = [];
+  var newColdIds = [];
+
+  if (store.isDirty) {
+    newColdIds = store.getAllIds();
+  }
+  else {
+    newHotIds = store.getChangedIds().filter(id => store.get(id) !== undefined);
+    newColdIds = store.sources.hot.filter(function getColdIds(geojson) {
+      return geojson.properties.id && newHotIds.indexOf(geojson.properties.id) === -1 && store.get(geojson.properties.id) !== undefined;
+    }).map(geojson => geojson.properties.id);
+  }
+
+  store.sources.hot = [];
+  let lastColdCount = store.sources.cold.length;
+  store.sources.cold = store.isDirty ? [] : store.sources.cold.filter(function saveColdFeatures(geojson) {
+    var id = geojson.properties.id || geojson.properties.parent;
+    return newHotIds.indexOf(id) === -1;
+  });
+
+  var coldChanged = lastColdCount !== store.sources.cold.length || newColdIds.length > 0;
+
+  newHotIds.concat(newColdIds).concat(newColdIds).map(function prepForViewUpdates(id) {
+    if (newHotIds.indexOf(id) > -1) {
+      return {source: 'hot', 'id': id};
     }
     else {
-      newHotIds = this.getChangedIds().filter(id => this.get(id) !== undefined);
-      newColdIds = this.sources.hot.filter(function getColdIds(geojson) {
-        return geojson.properties.id && newHotIds.indexOf(geojson.properties.id) === -1 && this.get(geojson.properties.id) !== undefined;
-      }.bind(this)).map(geojson => geojson.properties.id);
+      return {source: 'cold', 'id': id};
     }
+  }).forEach(function calculateViewUpdate(change) {
+    let {id, source} = change;
+    let feature = store.get(id);
+    let featureInternal = feature.internal(mode);
 
-    this.sources.hot = [];
-    let lastColdCount = this.sources.cold.length;
-    this.sources.cold = this.isDirty ? [] : this.sources.cold.filter(function saveColdFeatures(geojson) {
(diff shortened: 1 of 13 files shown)


davidtheclark:
Removed the `draw.mode.simple_select.selected.*` events, and updated documentation.

davidtheclark:
Was going to modify this so it emitted features, not featureIds, like the other events. Recalled @mcwhittemore's [comment](https://github.com/mapbox/mapbox-gl-draw/issues/365#issuecomment-226333423): "I don't think Draw should emit invalid geojson" ...

Are invalid features drawn and the map and selectable? Because if you can see and select an invalid feature, it would seem weird to me that the event would not return the invalid feature.

src/lib/util.js
```diff
@@ -205,6 +205,10 @@ function throttle(fn, time, context) {
   return wrapperFn;
 }
 
+module.exports.toDenseArray = function(x) {
```
mcwhittemore: while small, lets put this in its own file. I want to break `lib/util` up a bunch.

davidtheclark: Sounds good. How about I do this while writing unit tests?

mcwhittemore:
Invalid features will be present on the map during their `draw_*` mode, but if the feature is not made valid it will not stay on the map when the mode finishes.

Maybe `draw_*` shouldn't select anything so we don't need to worry about this?

The value of emitting information about a feature while it is being drawn is so that the application can do validation work of its own on the features shape.

src/lib/util.js
mcwhittemore: Sounds good

davidtheclark:
> Maybe draw_* shouldn't select anything so we don't need to worry about this?

I'll trust your call on this ... don't know enough yet about what is going on to feel like I have good judgment.

Is the only way to get into `direct_select` mode in Draw _from_ `simple_select` mode? If so, I’m wondering if we want to tailor the `select` and `deselect` events a little, so that when when you go from `simple_select` to `direct_select` the feature does not get deselected and re-selected. Instead, all other features get deselected, and the vertex gets added to the selection. What do you think?

Also: do you agree with the move to remove `draw.mode.simple_select.selected.*` events? Or do we still want those?

mcwhittemore:
> the vertex gets added to the selection

I haven't heard anyone ask for that so lets not expose vertex selection

> I’m wondering if we want to tailor the select and deselect events a little

Yea, In #365 I said:

> For the events we want to make sure we don't trash draw.unselected and draw.selected with the same ids. To do this, I think we want to the events to fire with the render cycle rather than when their state has changed.

I'm not sure if the render cycle is the right time, but its often and debounced already so it seems like a good enough spot.

davidtheclark:
> I'm not sure if the render cycle is the right time, but its often and debounced already so it seems like a good enough spot.

Ok, I will keep track of selected and deselected features, then fire the event only on render.

src/render.js
```diff
@@ -1,72 +1,86 @@
...
-      .filter(feature => feature !== undefined)
-      .filter(feature => feature.isValid())
-      .map(feature => feature.toGeoJSON());
+  const flushedSelectionSets = store.flushSeleted();
+  if (flushedSelectionSets.selected.length) {
+    store.ctxmap.fire('draw.select', { featureIds: flushedSelectionSets.selected });
+  }
+  if (flushedSelectionSets.deselected.length) {
+    store.ctxmap.fire('draw.deselect', { featureIds: flushedSelectionSets.deselected });
+  }
```
davidtheclark: @mcwhittemore I still have to update the tests, but can you look at this implementation, with `store.flushSelected`, and let me know if you think it will work?

mcwhittemore: I don't see a definition for `flushSeleted`

davidtheclark: https://github.com/mapbox/mapbox-gl-draw/pull/368/files/1cd2f63e2eabaa74f230cad147335133e9838223#diff-3bf35cd41da51153dcca3003656f7c67R192

I keep track of selected and deselected items since the last flush, then return them when you run that function.

davidtheclark:
I think this is starting to make sense to me. Here's a GIF showing select and deselect events getting fired:

selection-1

Here's some explanation of the times these events fire now:
- `select`
  - when feature is created
  - when feature is added to an existing selection with shift+click
  - when feature is contained in a box-selection
- `deselect`
  - when selected feature is deleted
  - when there is a click outside the selected feature, either on nothing or on another without using shift+click to add something else to its existing selection.

Does this seem right to you @mcwhittemore? If so I can write up some unit tests and polish the branch for merge.

davidtheclark:
Another note: nothing gets additionally selected when you move from `simple_select` to `direct_select` mode.

mcwhittemore:
@davidtheclark that sounds right to me. I'm not sure if I think `deselect` should be fired on `delete` though. Any idea what some other libs do in this situation?

davidtheclark:
> I'm not sure if I think deselect should be fired on delete though. Any idea what some other libs do in this situation?

Yeah I'm not sure either. I'll remove that for now in the interests of keeping the API lean. Will be easy to add in down the line if we decide it's important.

(I don't know what other libraries do ... I just spent a little while looking at other drawing libraries and couldn't quickly glean comparable behaviors.)

davidtheclark:
Ok, this branch is ready for review & possibly merge. It will be a breaking change so when we release we'll need to modify the dataset editor to listen for these new `select` and `deselect` events instead of `simple_select.selected`.

mcwhittemore:
I have two linestrings selected in simple_select. I click on a node of line a to enter direct select. line b becomes deselected, but there is no deselect event. I than click on the map to exit direct_select and get a `deselect` event with two feature ids.

davidtheclark:
Good discovery. I fixed that and another mode-changing bug in the recent commit. (If I had something selected then clicked a button to create a new feature, the previously selected feature lost selection but no deselect event fired.)

mcwhittemore:
@davidtheclark this looks great. I added a `getSelectedIds` function to the api so end users can take advantage of this work. If you want to look my commit over and than merge, that would be great.

Lets ship this thing.

davidtheclark:
Cool. Should I merge and (try to) publish a release or should we first try out this branch in studio?

davidtheclark:
Published in v0.10.0.

Issue #369, opened 2016-06-16
Add unit tests for feature types


davidtheclark:
I plan to work on this tomorrow, just wanted to log this issue so nobody else starts and our work overlaps.

PR #371, opened 2016-06-16
Add unit tests for features

davidtheclark:
Ref #369


test/feature.test.js
```diff
@@ -0,0 +1,120 @@
...
+test('Feature contrusctor and API', t => {
+  const featureGeoJson = createFeature('line');
+  const ctx = createMockCtx();
+  const feature = new Feature(ctx, featureGeoJson);
+
+  // Instance members
+  t.equal(feature.ctx, ctx, 'feature.ctx');
+  t.equal(feature.coordinates, featureGeoJson.geometry.coordinates, 'feature.coordinates');
+  t.equal(feature.properties, featureGeoJson.properties, 'feature.properties');
+  t.equal(feature.atLastRender, null, 'feature.atLastRender starts with null');
```
mcwhittemore: I don't think we use this anywhere anymore

mcwhittemore:
This is looking really good.

davidtheclark:
Ready for review and merge. @mcwhittemore

src/feature_types/polygon.js
```diff
@@ -14,19 +9,17 @@ var Polygon = function(ctx, geojson) {
...
 Polygon.prototype.isValid = function() {
-  return this.coordinates.every(function(ring) {
-    return ring.length > 2;
-  });
+  return this.coordinates.every(ring => ring.length > 2);
 };
 
 Polygon.prototype.incomingCoords = function(coords) {
-  if (!firstMatchesLast(coords)) throw new Error('First and last positions must be equivalent. Use setCoordinates method.');
+  if (!linearRingsAreClosed(coords)) throw new Error('First and last positions must be equivalent. Use setCoordinates method.');
```
mcwhittemore: Lets drop these error message. We've decided they were a bad idea.

src/feature_types/polygon.js
```diff
@@ -14,19 +9,17 @@ var Polygon = function(ctx, geojson) {
...
   this.coordinates = coords.map(ring => ring.slice(0, -1));
   this.changed();
 };
 
 Polygon.prototype.setCoordinates = function(coords) {
-  if (firstMatchesLast(coords)) throw new Error('First and last positions must not be equivalent. Use incomingCoords method.');
+  if (linearRingsAreClosed(coords)) throw new Error('First and last positions must not be equivalent. Use incomingCoords method.');
```
mcwhittemore: Drop this `throw` as well.

src/feature_types/polygon.js
davidtheclark: So just don't validate the rings? Or do you want something else to happen?

src/lib/linear_rings_are_closed.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Returns a Boolean indicating whether all the linearRings
+ * in an array of them, as in the coordinates of a Polygon,
+ * are closed.
+ *
+ * @param {Array<Array<Array<number>>>} rings
+ * @return {Boolean} `true` if all rings are closed.
+ */
+function linearRingsAreClosed(rings) {
```
mcwhittemore: I think this is just used to throw errors. Might be useful in tests to make sure nothing is screwing this up, but throwing this error isn't problematic as GeoJSON doesn't say this can't happen.

mcwhittemore:
@davidtheclark this is looking great. Take my comments above with a grain of salt (this PR is to good to be stopped by more clean up work). Merge when you're ready.

davidtheclark:
Removed the throws and their tests. Will merge on green light.

Issue #370, opened 2016-06-16
Add unit tests for lib files


davidtheclark:
I plan to do this soon, just wanted to file issue so work doesn't overlap.

Is there a rationale distinguishing the functions in `lib/` with their own file and the functions exported by `lib/util.js`? I'd like to know so that I add things to the right place, if I were to add things.

mcwhittemore:
@davidtheclark `lib/util` was partial taken from Leaflet I think. Please feel free to break it up.

mcwhittemore:
The only thing really left here is the polyfill... and a few edge cases in mode_handler and StringSet.

![image](https://cloud.githubusercontent.com/assets/1551510/16421095/d20b4ece-3d21-11e6-8661-0a6dd2346816.png)

davidtheclark:
Addressed the polyfills lack with #427, and addressing mode_handler with #436.

PR #374, opened 2016-06-17
Lib tests

davidtheclark:
Ref #370.


davidtheclark:
@samanpwbb or @mcwhittemore: I think the only thing in `lib/` that still needs tests is `mode_handler`. Rather than prolonging this PR further I'll put that in another. Can we review this one and possibly merge?

Definitely needs a double-check in the UI that untested functionality is not broken, because I did some refactoring — mostly just splitting up and renaming files. Seems to me to work the same way as before on the debug page.

davidtheclark:
I'll try to write down some of those splittings/renamings so it makes the diff easier to read:

- `toVertex` -> `createVertex` and `toMidpoint` -> `createMidpoint`
- All `utils.js` split out. A few of them were not used anywhere, so I deleted them.
- From `featuresAt()` I extracted `mapEventToBoundingBox()` and `sortFeatures()`

I think that's it.

src/lib/common_selectors.js
```diff
@@ -30,7 +31,7 @@ module.exports = {
     return e.featureTarget !== undefined && e.featureTarget.properties.meta === 'feature';
   },
   isShiftDown: function(e) {
-    return e.originalEvent.shiftKey === true;
+    return e.originalEvent && e.originalEvent.shiftKey === true;
```
mcwhittemore: This will return undefined if `originalEvent` is undefined and not false.

src/lib/create_control_button.js
```diff
@@ -0,0 +1,41 @@
...
+  button.addEventListener('click', (e) => {
+    e.preventDefault();
+    e.stopPropagation();
+
+    const el = e.target;
+
+    if (el.classList.contains('active')) {
+      el.classList.remove('active');
+    } else {
+      domUtils.removeClass(document.querySelectorAll(`.${Constants.CCONTROL_BUTTON_CLASS}`), 'active');
```
mcwhittemore: One too many `C`s.

src/lib/sort_features.js
```diff
@@ -0,0 +1,37 @@
...
+
+  if (score === 0 && a.geometry.type === 'Polygon') {
+    return a.area - b.area;
+  }
+
+  return score;
+}
+
+// Sort in the order above, then sort polygons by area ascending.
+function sortFeatures(features) {
```
mcwhittemore: we don't need the `const` here. We could just `map`, `sort`, `map`, `return`.

mcwhittemore:
@davidtheclark this is looking great. It seems to work well, though I did find a small bug with the `control class`.

After this merges, I'd like to get some coverage reports so that we can see where we are missing coverage.

davidtheclark:
Thanks for the feedback @mcwhittemore. I will make changes accordingly.

davidtheclark:
@mcwhittemore Pushed changes based on the feedback. I'll create an issue for test coverage, and start another PR with some more tests.

mcwhittemore:
Sounds good!

PR #376, opened 2016-06-20
simple_select tests

samanpwbb:
Not ready. Not able to select or move things by firing events on the map.


samanpwbb:
Problem comes down to featuresAt always failing to return anything.

samanpwbb:
🙅

PR #381, opened 2016-06-21
Add ui tests

davidtheclark:
In addition to writing the `ui` module's tests, I renamed a few of its functions, removed some now-unnecessary stuff, and moved more strings into `constants.js` — which revealed that there is some inconsistency in the class names (mapboxgl or mapbox-gl) that we should clear up at some point.

cc @mcwhittemore


src/ui.js
```diff
@@ -1,169 +1,164 @@
...
       }
-    };
+      delete buttonElements[buttonId];
+    });
+  }
+
+  return {
+    setButtonActive,
+    deactivateButtons,
+    queueContainerClasses,
```
mcwhittemore: Container is a bit of an overloaded word in Draw. I think there are two containers right now (one for the full map and one for where the controls go). May `queueMapClasses` would be better?

davidtheclark: Sure! If the class are applied to the map's containing element, that seems fine to me.

mcwhittemore:
@davidtheclark lets do that small rename and than merge this.

PR #382, opened 2016-06-21
Simple select

### LINKED ISSUE #383: Impossible to drag a multipolygon
dragging a multipolygon results in this error:

screen shot 2016-06-21 at 7 51 43 pm

Here: https://github.com/mapbox/mapbox-gl-draw/blob/36f3bced9ae73844ac8c79312e063988e2f1a3d0/src/modes/simple_select.js#L214

mcwhittemore:
I was able to get this up to 60% today. Most of the time was spent [mocking query rendered features](https://github.com/mapbox/mapbox-gl-js-mock/pull/5). I think I'll be at ~100 tomorrow.

**before**
image

**after**
image

Closes #383

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,9 @@
     "test": "npm run lint && npm run tape",
     "lint": "eslint --no-eslintrc -c .eslintrc index.js src",
     "tape": "tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
+    "tss": "tape -r ./test/mock-browser.js -r babel-register test/simple_select.test.js",
     "coverage": "NODE_ENV=test nyc --reporter html npm run tape && opener coverage/index.html",
+    "css": "NODE_ENV=test nyc --reporter html npm run tss && opener coverage/index.html",
     "build": "NODE_ENV=production browserify index.js > dist/mapbox-gl-draw.js",
     "prepublish": "NODE_ENV=production browserify index.js | uglifyjs -c -m > dist/mapbox-gl-draw.js",
     "start": "node server.js"
@@ -72,6 +74,7 @@
     "point-geometry": "0.0.0",
     "ramda": "^0.17.1",
     "sinon": "^1.17.4",
-    "turf-area": "^1.1.1"
+    "turf-area": "^1.1.1",
+    "webgl-mock": "^0.1.6"
   }
 }
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "express": "^4.13.4",
     "geojson-validation": "^0.1.6",
     "mapbox-gl": "^0.19.0",
-    "mapbox-gl-js-mock": "^0.22.0",
+    "mapbox-gl-js-mock": "^0.23.0",
     "mock-browser": "^0.92.10",
     "nyc": "^6.6.1",
     "opener": "^1.4.1",
--- a/package.json
+++ b/package.json
@@ -15,9 +15,7 @@
     "test": "npm run lint && npm run tape",
     "lint": "eslint --no-eslintrc -c .eslintrc index.js src",
     "tape": "tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
-    "tss": "tape -r ./test/mock-browser.js -r babel-register test/simple_select.test.js",
     "coverage": "NODE_ENV=test nyc --reporter html npm run tape && opener coverage/index.html",
-    "css": "NODE_ENV=test nyc --reporter html npm run tss && opener coverage/index.html",
     "build": "NODE_ENV=production browserify index.js > dist/mapbox-gl-draw.js",
     "prepublish": "NODE_ENV=production browserify index.js | uglifyjs -c -m > dist/mapbox-gl-draw.js",
     "start": "node server.js"
@@ -72,9 +70,7 @@
     "geojsonhint": "^1.2.0",
     "hat": "0.0.3",
     "point-geometry": "0.0.0",
-    "ramda": "^0.17.1",
     "sinon": "^1.17.4",
-    "turf-area": "^1.1.1",
-    "webgl-mock": "^0.1.6"
+    "turf-area": "^1.1.1"
   }
 }
--- a/src/render.js
+++ b/src/render.js
@@ -76,7 +76,7 @@ module.exports = function render() {
     store.ctx.map.fire('draw.deselect', { featureIds: flushedSelectionSets.deselected });
   }
 
-  store.ctx.map.fire('draw.render');
+  store.ctx.map.fire('draw.render', {});
   cleanup();
 
   function cleanup() {
--- a/test/simple_select.test.js
+++ b/test/simple_select.test.js
@@ -86,7 +86,7 @@ test('simple_select', t => {
     });
   });
 
-  t.test('simple_select - unselect', t => {
+  t.test('simple_select - deselect', t => {
     var id = Draw.add(cloneFeature('point'));
     Draw.changeMode('simple_select', [id]);
     map.fire.reset();
@@ -99,15 +99,15 @@ test('simple_select', t => {
         var args = getFireArgs().filter(arg => arg[0] === 'draw.deselect');
         t.equal(args.length, 1, 'should have one and only one selected end event');
         if (args.length > 0) {
-          t.equal(args[0][1].featureIds.length, 1, 'should unselect only one feautre');
-          t.equal(args[0][1].featureIds[0], id, 'should unselect the feature we expect it to select');
+          t.equal(args[0][1].featureIds.length, 1, 'should deselect only one feautre');
+          t.equal(args[0][1].featureIds[0], id, 'should deselect the feature we expect it to select');
         }
         cleanUp(() => t.end());
       });
     });
   });
 
-  t.test('simple_select - click on an unselected feature', t => {
+  t.test('simple_select - click on an deselected feature', t => {
     var id = Draw.add(cloneFeature('polygon'));
     console.log(JSON.stringify(cloneFeature('polygon')));
     Draw.changeMode('simple_select');
@@ -144,8 +144,8 @@ test('simple_select', t => {
         args = args.filter(arg => arg[0] === 'draw.deselect');
         t.equal(args.length, 1, 'should have one and only one selected end event');
         if (args.length > 0) {
-          t.equal(args[0][1].featureIds.length, 1, 'should select only one feautre');
-          t.equal(args[0][1].featureIds[0], id, 'should select the feature we expect it to select');
+          t.equal(args[0][1].featureIds.length, 1, 'should deselect only one feautre');
+          t.equal(args[0][1].featureIds[0], id, 'should deselect the feature we expect it to deselect');
         }
         cleanUp(() => t.end());
       });
@@ -215,7 +215,7 @@ test('simple_select', t => {
     });
   });
 
-  t.test('simple_select - click on an unselected feature with shift down while having another feature selected', t => {
+  t.test('simple_select - click on an deselected feature with shift down while having another feature selected', t => {
     var pointId = Draw.add(cloneFeature('point'));
     var id = Draw.add(cloneFeature('polygon'));
     Draw.changeMode('simple_select', [pointId]);
@@ -240,7 +240,7 @@ test('simple_select', t => {
     });
   });
 
-  t.test('simple_select - click on an unselected feature with shift up, while having another feature selected', t => {
+  t.test('simple_select - click on an deselected feature with shift up, while having another feature selected', t => {
     var pointId = Draw.add(cloneFeature('point'));
     var id = Draw.add(cloneFeature('polygon'));
     Draw.changeMode('simple_select', [pointId]);
--- a/test/simple_select.test.js
+++ b/test/simple_select.test.js
@@ -11,6 +11,10 @@ test('simple_select', t => {
 
   var map = createMap();
   spy(map, 'fire');
+  spy(map.doubleClickZoom, 'enable');
+  spy(map.doubleClickZoom, 'disable');
+  spy(map.dragPan, 'enable');
+  spy(map.dragPan, 'disable');
 
   var Draw = GLDraw();
   map.addControl(Draw);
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -71,6 +73,7 @@
     "point-geometry": "0.0.0",
     "ramda": "^0.17.1",
     "sinon": "^1.17.4",
-    "turf-area": "^1.1.1"
+    "turf-area": "^1.1.1",
+    "webgl-mock": "^0.1.6"
```
mcwhittemore: - [ ] remove this

package.json
```diff
@@ -15,7 +15,9 @@
     "test": "npm run lint && npm run tape",
     "lint": "eslint --no-eslintrc -c .eslintrc index.js src",
     "tape": "tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
+    "tss": "tape -r ./test/mock-browser.js -r babel-register test/simple_select.test.js",
     "coverage": "NODE_ENV=test nyc --reporter html npm run tape && opener coverage/index.html",
+    "css": "NODE_ENV=test nyc --reporter html npm run tss && opener coverage/index.html",
```
mcwhittemore: - [ ] remove this

package.json
```diff
@@ -15,7 +15,9 @@
     "test": "npm run lint && npm run tape",
     "lint": "eslint --no-eslintrc -c .eslintrc index.js src",
     "tape": "tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
+    "tss": "tape -r ./test/mock-browser.js -r babel-register test/simple_select.test.js",
```
mcwhittemore: - [ ] remove this

test/simple_select.test.js
```diff
@@ -0,0 +1,160 @@
...
+      stopPropagation: function() {},
+      button: 0,
+      clientX: lng,
+      clientY: lat
+    },
+    point: {x: lng, y:lat},
+    lngLat: {lng: lng, lat: lat}
+  };
+
+  e.featureTarget = Draw.getFeatureIdsAt(lng, lat)[0];
```
mcwhittemore: - [ ] I shouldn't need this.

test/simple_select.test.js
```diff
@@ -0,0 +1,160 @@
...
+var feature = features.point;
+
+var map = createMap();
+spy(map, 'fire');
+
+var Draw = GLDraw();
+map.addControl(Draw);
+
+var afterNextRender = function(cb) {
+  setTimeout(cb, 32);
```
mcwhittemore: - [ ] have draw emit a `rendered` event and use this for the next tick.

test/simple_select.test.js
```diff
@@ -0,0 +1,160 @@
...
+
+var getFireArgs = function() {
+  var args = [];
+  for (var i=0; i<map.fire.callCount; i++) {
+    args.push(map.fire.getCall(i).args);
+  }
+  return args;
+}
+
+var makeMouseEvent = function(lng, lat, shiftKey = false) {
```
davidtheclark: It's be great to extract this out into a `test_util`. We'll need it in other modes.

mcwhittemore:
98.82%, I'm going to start cleaning up the code now.

image

mcwhittemore:
@davidtheclark ready for your review

src/render.js
```diff
@@ -76,7 +76,7 @@ module.exports = function render() {
     store.ctx.map.fire('draw.deselect', { featureIds: flushedSelectionSets.deselected });
   }
 
-  store.ctx.map.fire('draw.render');
+  store.ctx.map.fire('draw.render', {});
```
davidtheclark: This event is undocumented ... is it being used in studio? Should we keep it and document it?

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+
+  var getFireArgs = function() {
+    var args = [];
+    for (var i=0; i<map.fire.callCount; i++) {
+      args.push(map.fire.getCall(i).args);
+    }
+    return args;
+  }
+
+  t.test('simple_select - setup', t => {
```
davidtheclark: How does this test the setup of `simple_select`?

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+      if (donedone === false) {
+        t.pass('map loaded');
+        t.end();
+        donedone = true;
+      }
+    };
+    if (map.loaded()) {
+      done();
+    }
+    map.on('load', done);
```
davidtheclark: I don't understand why `donedone` is necessary above ... Could you not `return`, like `if (map.loaded()) return done()`?

src/render.js
mcwhittemore: Just added it to the docs.

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+  });
+
+
+  t.test('simple_select box select', t => {
+    Draw.add(cloneFeature('negitivePoint'));
+    var id = Draw.add(cloneFeature('point'));
```
davidtheclark: I think it would make the test more transparent to specify the coordinates within this function. Right now we're trusting that the point you're cloning is within the box you specify below; but it could be made explicit.

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+    afterNextRender(() => {
+      map.fire('mousedown', makeMouseEvent(0, 0, true));
+      map.fire('mousemove', makeMouseEvent(7, 7, true));
+      map.fire('mousemove', makeMouseEvent(15, 15, true));
+      map.fire('mouseup', makeMouseEvent(15, 15, true));
+
+      afterNextRender(() => {
+        var args = getFireArgs().filter(arg => arg[0] === 'draw.select');
+        t.equal(args.length, 1, 'should have one and only one selected start event');
+        if (args.length > 0) {
```
davidtheclark: Why use this conditional if you just asserted the length is 1?

test/simple_select.test.js
mcwhittemore: Ah, bad name. The goal of this is just to confirm that that map is ready. Pushed an update to make that more clear and do this in a better way.

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+      });
+    });
+  });
+
+  t.test('simple_select - box select over 1001 points', t => {
+
+    for(var i=0; i<1001; i++) {
+      Draw.add(cloneFeature('point'));
+    }
+    map.fire.reset();
```
davidtheclark: Probably want to move at least a couple of those points to different coordinates so the tests checks that, too.

test/simple_select.test.js
mcwhittemore: yea, I just had this thought too. Updated.

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+
+    afterNextRender(() => {
+      map.fire('mousedown', makeMouseEvent(0, 0, true));
+      map.fire('mousemove', makeMouseEvent(7, 7, true));
+      map.fire('mousemove', makeMouseEvent(15, 15, true));
+      map.fire('mouseup', makeMouseEvent(15, 15, true));
+
+      afterNextRender(() => {
+        t.equal(getFireArgs().filter(arg => arg[0] === 'draw.select').length, 0, 'there should be no draw.select event');
```
davidtheclark: Wait ... should it have selected 1001 points?

test/simple_select.test.js
```diff
@@ -0,0 +1,348 @@
...
+        cleanUp(() => t.end());
+      });
+    });
+  });
+
+  t.test('simple_select - unselect', t => {
+    var id = Draw.add(cloneFeature('point'));
+    Draw.changeMode('simple_select', [id]);
+    map.fire.reset();
```
davidtheclark: All these `map.fire.resets()` at the beginning of the case are unnecessary, right? Because of `cleanup()`?

test/simple_select.test.js
mcwhittemore: `tape` doesn't stop a test suite when asserts fail, so the `[0][1]` will throw an error which forces tape to exit.

test/simple_select.test.js
mcwhittemore: There is a check in `simple_select` to not select the features if its going to select more than 1000... I'll open an issue talking about that because I think its weird too.

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+
+    afterNextRender(() => {
+      map.fire.reset();
+      map.fire('mousedown', makeMouseEvent(50, 30, true));
+      map.fire('mouseup', makeMouseEvent(50, 30, true));
+
+      afterNextRender(() => {
+        var args = getFireArgs();
+        args = args.filter(arg => arg[0] === 'draw.deselect');
+        t.equal(args.length, 1, 'should have one and only one selected end event');
```
davidtheclark: How can this deselect a feature if no feature has yet been selected?

test/simple_select.test.js
davidtheclark: Oh, so you're testing that. In that case, probably need a test that succeeds in selecting multiple features with a box-select, right?

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+        if (args.length > 0) {
+          t.equal(args[0][1].featureIds.length, 1, 'should select only one feautre');
```
davidtheclark: The descriptions should say `deselect`

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+    var args = getFireArgs();
+    t.equal(args.length, 1, 'should have one and only one event');
+    t.equal(args[0][0], 'draw.deleted');
+    t.equal(args[0][1].featureIds.length, 1, 'should delete only one feautre');
+    t.equal(args[0][1].featureIds[0].id, id, 'should delete the feature we expect it to delete');
+
+    var selectedFeatures = Draw.getSelectedIds();
+    t.equal(selectedFeatures.length, 0, 'nothing should be selected anymore');
+    cleanUp(() => t.end());
+  });
```
davidtheclark: Should we have different tests for deletion with `Draw.trash()` and deletion using the keyboard?

test/simple_select.test.js
mcwhittemore: It also removes anything that happened because of adding, changing mode, so on. Some could go, some might need to stay.

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+
+  t.test('simple_select - click on an selected feature with shift up to enter direct_select', t => {
+    Draw.deleteAll();
+    var id = Draw.add(cloneFeature('polygon'));
+    Draw.changeMode('simple_select', [id]);
+
+    afterNextRender(() => {
+      map.fire.reset();
+      map.fire('mousedown', makeMouseEvent(50, 30, false));
+      map.fire('mouseup', makeMouseEvent(50, 30, false));
```
davidtheclark: Curious: is there a reason you're using mousedown/up instead of the `click()` function?

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+
+      afterNextRender(() => {
+        var args = getFireArgs();
+        args = args.filter(arg => arg[0] === 'draw.modechange');
+        t.equal(args.length, 1, 'should have one and only one selected end event');
```
davidtheclark: mode change event

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+
+    afterNextRender(() => {
+      map.fire.reset();
+      map.fire('mousedown', makeMouseEvent(clickPosition[0], clickPosition[1]));
+      map.fire('mouseup', makeMouseEvent(clickPosition[0], clickPosition[1]));
+
+      afterNextRender(() => {
+        var args = getFireArgs();
+        args = args.filter(arg => arg[0] === 'draw.modechange');
+        t.equal(args.length, 1, 'should have one and only one selected end event');
```
davidtheclark: mode change event

test/simple_select.test.js
mcwhittemore: Pretty sure they are the same thing. We should have tests that confirm that delete emits a trash event maybe?

test/simple_select.test.js
```diff
@@ -0,0 +1,346 @@
...
+
+      var movedPolygon =Draw.get(polygonId);
+      t.equal(movedPolygon.geometry.coordinates[0][0][0], cloneFeature('polygon').geometry.coordinates[0][0][0] + 25, 'polygon lng moved');
+      t.equal(movedPolygon.geometry.coordinates[0][0][1], cloneFeature('polygon').geometry.coordinates[0][0][1] + 25, 'polygon lat moved');
+      t.equal(countPositions(movedPolygon), countPositions(cloneFeature('polygon')), 'polygon has same number of postions');
+
+      var movedMultiPolygon =Draw.get(multiPolygonId);
+      t.equal(movedMultiPolygon.geometry.coordinates[0][0][0][0], cloneFeature('multiPolygon').geometry.coordinates[0][0][0][0] + 25, 'multiPolygon lng moved');
+      t.equal(movedMultiPolygon.geometry.coordinates[0][0][0][1], cloneFeature('multiPolygon').geometry.coordinates[0][0][0][1] + 25, 'multiPolygon lat moved');
+      t.equal(countPositions(movedMultiPolygon), countPositions(cloneFeature('multiPolygon')), 'multiPolygon has same number of postions');
```
davidtheclark: Oh did you fix #383 while doing this?

test/simple_select.test.js
davidtheclark: Yeah ... good to know that you prefer this way instead of having the error thrown. I will keep that in mind.

test/simple_select.test.js
mcwhittemore: A bunch of the tests have dragging and I didn't want to add dragging without refactoring out click and it felt weird to sometimes fire the events and sometimes have a helper doing it. I'll refactor click out and make a drag helper on direct_select. I was just worried we'd clash if I worked too much in those files.

test/simple_select.test.js
```diff
@@ -0,0 +1,326 @@
...
+    afterNextRender(() => {
+      map.fire.reset();
+      map.fire('mousedown', makeMouseEvent(50, 30, false));
+      map.fire('mouseup', makeMouseEvent(50, 30, false));
+
+      afterNextRender(() => {
+        t.equal(Draw.getSelectedIds().indexOf(pointId) === -1, true, 'point is no longer selected');
+        t.equal(Draw.getSelectedIds().indexOf(id) != -1, true, 'polygon is now selected');
+        var args = getFireArgs();
+        args = args.filter(arg => arg[0] === 'draw.select');
```
davidtheclark: Should check if the deselect event fired, too.

test/simple_select.test.js
mcwhittemore: Yep. Didn't realize there was an issue for it. Will add it to the PR so it closes when merged.

test/simple_select.test.js
mcwhittemore: part of me just wants to move to mocha but that is a different convo and if/when it happens it shouldn't happen now.

test/simple_select.test.js
mcwhittemore: Hmmm... coverage doesn't need it, but its probs a good test. Adding it now.

test/simple_select.test.js
davidtheclark: Oh, I guess the backspace functionality is not specific to this mode, so nevermind.

davidtheclark:
That's a lot to test, and a lot of browser/draw problems you dealt with!

There may be a few more to write, though. I think the coverage report is saying that certain lines are covered but the tests aren't yet ensuring that those lines are doing the right thing. The lines could have typos or do something else wrong and the tests wouldn't fail. I think it's pretty much just invocations on something in `ctx`, e.g. 

- https://github.com/mapbox/mapbox-gl-draw/blob/65455365a2a77ab0a357e747d6273d6c002ba945/src/modes/simple_select.js#L69
- https://github.com/mapbox/mapbox-gl-draw/blob/65455365a2a77ab0a357e747d6273d6c002ba945/src/modes/simple_select.js#L123
- https://github.com/mapbox/mapbox-gl-draw/blob/65455365a2a77ab0a357e747d6273d6c002ba945/src/modes/simple_select.js#L51

test/simple_select.test.js
davidtheclark: It's probably not more than an hour or two of busywork, so if you think it will make you more productive it may be worth it even now ...

mcwhittemore:
@davidtheclark the ui one makes sense but the other two are invoking gl-js functions. Coverage tells me the function has been invoked so a spy on it seems weird. Is this what you are expecting?

davidtheclark:
> Coverage tells me the function has been invoked

But couldn't the code be passing the wrong argument?

davidtheclark:
Also, couldn't somebody change that line of code, e.g. to a `console.log()`, and tests wouldn't fail (and coverage wouldn't decrease)?

mcwhittemore:
@davidtheclark you achieve this by mocking the context in the `draw_*` tests. I don't want to mock the context as the rendering pipeline plays a large part in how some of these tests work. Thoughts?

davidtheclark:
I don't know what best practices are (is there someone we should ask who might know?), but here are two ideas:

- We could expose the context object in some way for testing purposes only, so you could spy on functions —— e.g. `Draw.__internal`
- Pull some of those things out into standalone functions that get passed a context object. Then you could test them independently with a mock context object.

davidtheclark:
> We could expose the context object in some way for testing purposes only, so you could spy on functions —— e.g. Draw.__internal

That option would be useful for writing a whole bunch of event tests ...

mcwhittemore:
> We could expose the context object in some way for testing purposes only

`mapbox-gl-js` does this I think. I'll look at how they are doing it.

mcwhittemore:
@davidtheclark ready for your review.

davidtheclark:
Latest commit looks great. I say merge when you feel the time is right.

PR #384, opened 2016-06-22
Draw line string mode

davidtheclark:
This is branched off #381, so can't be merged (and diff won't make a lot of sense) until that is merged.

@mcwhittemore I took some ideas from #382 for testing interactions. Tomorrow let's talk about these two PRs and get on the same page about the patterns in them? One thing I wasn't sure how to do was test `keyup` events ...

PR #388, opened 2016-06-22
draw_line_string mode tests

davidtheclark:
Continuation of #384

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "nyc": "^6.6.1",
     "opener": "^1.4.1",
     "sinon": "^1.17.4",
+    "synthetic-dom-events": "^0.3.0",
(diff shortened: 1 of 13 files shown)


davidtheclark:
@mcwhittemore ready for review

mcwhittemore:
Can you add a test where you start drawing and than exit before a click. This should get this mode to 100%.

image

mcwhittemore:
This works well. Once you add that last test, lets merge.

davidtheclark:
@mcwhittemore I think I was able to just delete that line, because the deletion of incomplete elements is already done in https://github.com/mapbox/mapbox-gl-draw/pull/388/files#diff-7a0124268c46d39459f7595fd9d4d1caR75 as far as I can tell.

PR #389, opened 2016-06-22
adding in a few simple test utils

mcwhittemore:
@davidtheclark for the review

I tried to do get the afterNextRender to work right, but the spy is in some weird state and throwing errors. 32ms timeout will do for this PR.


davidtheclark:
Right now we have a bunch of utils in `test_utils.js`. I guess this PR can be the start of splitting those up?

If so, seems good to me. Merge away.

mcwhittemore:
yea, I think we should break that file down too. Move the fixtures to the fixtures folder and the utils here.

PR #392, opened 2016-06-22
Fix select event docs; closes #391

### LINKED ISSUE #391: draw.deselect documentation is inaccurate
https://github.com/mapbox/mapbox-gl-draw/blob/master/API.md#drawdeselect

> Fired when a feature is deselected. The payload is the GeoJSON of the feature that was deselected.

The actual payload is a featureIds array https://github.com/mapbox/mapbox-gl-draw/blob/3b191e50f2cfd8ada99a64191d202ee864757ac5/src/render.js#L76

davidtheclark:
Ref #391.

@tmcw for review.

### DIFF
--- a/API.md
+++ b/API.md
@@ -248,7 +248,7 @@ This is fired when feature coordinates are changed, and there is a mouse click a
 
 ### draw.select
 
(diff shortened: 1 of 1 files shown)


mcwhittemore:
@davidtheclark Looks good. Merging.

Issue #395, opened 2016-06-22
`draw.update` to replace `draw.modified` and draw.changed`


mcwhittemore:
Currently we have two events that emit updates that happen within Draw. `draw.changed` emits with every frame and `draw.modified` updates with every mouse click.

@davidtheclark is currently working on making `draw.modified` a bit clearer [here](https://github.com/mapbox/mapbox-gl-draw/issues/393).

Once that has merged I think `draw.modified` will be much better than `draw.changed` and (as @davidtheclark pointed out) it could have a better name.

Is there any reason we should keep the current functionality of `draw.changed` or can we remove this and go with `draw.modified` or `draw.update`?

mcwhittemore:
Replaced by #397

PR #396, opened 2016-06-22
Change modified event firing; closes #393

### LINKED ISSUE #393: The first time you click any feature a draw.modified event is fired
And it should not be, since you are not modifying features.

davidtheclark:
Refs #393 and #395.

This seemed to me like a straightforward way to make the `draw.modified` event more meaningful, and it addresses all the cases in https://github.com/mapbox/mapbox-gl-draw/issues/393#issuecomment-227880372: it *only* fires: 

- When you finish moving a feature.
- When you finish moving a vertex. 
- When you delete a vertex.

Tests are failing because there are some `draw.modified` event tests that want the event to fire a lot more often than it would after this modification.

What do you think @mcwhittemore , @ghoshkaj , @tmcw ?

### DIFF
--- a/src/modes/direct_select.js
+++ b/src/modes/direct_select.js
@@ -43,6 +43,12 @@ module.exports = function(ctx, opts) {
     numCoords = coordPos.length;
   };
 
+  var emitModify = () => {
+    ctx.map.fire('draw.modified', {
+      features: [feature]
+    });
+  };
+
   return {
     start: function() {
       ctx.store.setSelected(featureId);
@@ -67,6 +73,7 @@ module.exports = function(ctx, opts) {
         }
       });
       this.on('mouseup', () => true, function() {
+        if (dragging) emitModify();
         dragging = false;
         coordPos = null;
         numCoords = null;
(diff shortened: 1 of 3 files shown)


src/modes/direct_select.js
```diff
@@ -67,6 +73,7 @@ module.exports = function(ctx, opts) {
         }
       });
       this.on('mouseup', () => true, function() {
+        if (dragging) emitModify();
```
mcwhittemore: What if I haven't moved the mouse at all? Just clicked down, waited a second, released?

src/modes/simple_select.js
```diff
@@ -69,6 +69,10 @@ module.exports = function(ctx, startingSelectedIds) {
     });
   };
 
+  var emitModify = function() {
+    ctx.map.fire('draw.modified', { features });
```
mcwhittemore: `features` should be dropped from this file and we should use what is currently selected.

src/modes/direct_select.js
mcwhittemore: So, `dragging` doesn't mean we've moved. We might want to rename this var.

src/modes/simple_select.js
```diff
@@ -168,6 +172,7 @@ module.exports = function(ctx, startingSelectedIds) {
       });
 
       this.on('mouseup', () => true, function(e) {
+        if (dragging) emitModify();
```
mcwhittemore: `dragging` doesn't mean we've dragged, just that we can drag. We might want to rename this to `canDrag`.

mcwhittemore:
@davidtheclark if we do this to `draw.modified` we'd also need some sort of event to inform about features having been created.

davidtheclark:
Closing in favor of #407

PR #401, opened 2016-06-23
Add draw_polygon tests

davidtheclark:


mcwhittemore:
@davidtheclark is this ready for review?

davidtheclark:
@mcwhittemore No, I abandoned it part way to work on the events. Will pick it up again after the events are done.

davidtheclark:
@mcwhittemore now this is ready for review.

test/draw_polygon.test.js
```diff
@@ -0,0 +1,428 @@
...
+    },
+    geometry: {
+      type: 'LineString',
+      coordinates: [[0, 0], [0, 10]]
+    }
+  }, 'a line string is sent to the callback');
+  t.end();
+});
+
+test('draw_polygon render, inactive, with only two vertices', t => {
```
mcwhittemore: This is inactive because there is no `id` not because the feature only has two vertices.

davidtheclark: Do you think we should remove this test? Or leave as-is?

mcwhittemore: it looks like its tested again lower down, so lets drop this.

PR #406, opened 2016-06-23
Get feature ids at point

mcwhittemore:
@tmcw for review.


davidtheclark:
Since Tom is gone, I'll review? I think it looks good to go.

Issue #397, opened 2016-06-23
Events re-vamp


davidtheclark:
I think that it may be worth our while to think through and re-vamp all of the events that get emitted by Draw. We could tackle #395, #393, #367, #316, #271, #307, #385, and other undocumented confusions all at once in a little overhaul. 

Issues, questions, and debugging difficulties keep coming up as a result of one of the following:

- An event that seems like it should exist not existing (e.g. add/create).
- Events are firing in unintuitive or unclear ways.

*I think we need to more clearly define our intentions, in documentation and tests, about exactly when each event is going to fire*. The event timing should make intuitive sense, and should be useful for Studio.

I'm going to draft proposed docs for re-vamped events below, and if you think this isn't a silly effort we can alter it according to your feedback, then I'll work on a PR to make it happen.

**Those of you with more experience trying to use Draw's events in Studio ( @mcwhittemore , @samanpwbb , @ghoshkaj , @tmcw ) please chime in about whether the proposal below would fulfill the needs and make things better** (or if it's not worth trouble).

---

I'm thinking maybe all these events should fire *only as a result of user interaction tracked by Draw* — i.e. not when Draw's API is used to programmatically do things like change modes, or add and delete features. (Which would avoid weirdness like this: https://github.com/mapbox/studio/blob/dev-pages/src/datasets_components/editor/index.js#L214)

### `draw.selectionchange`

Fired when the selection is changed (i.e. something is selected or deselected) as the result of a user interaction. The following interactions will trigger this event:

- Click on a feature to select it.
- Create a box-selection that includes at least one feature.
- When a feature is already selected, shift-click on another feature to add it to the selection.
- Click outside the selected feature(s) to deselect.
- Finish creating a feature.

The event data is an object with the following shape:

```js
{
  // Array of GeoJSON objects representing the features 
  // that are selected, after the change
  features: [{..}, {..}, ..]
}
```

### `draw.create`

Fired when a feature is created as the result of a user interaction. The following interactions will trigger this event:

- Finish drawing a feature. Simply clicking will create a Point. A LineString or Polygon is only created when the user has finished drawing it — i.e. double-clicked the last vertex or hit Enter — and the drawn feature is valid.

The event data is an object with the following shape:

```js
{
  // Array of GeoJSON objects representing the features that were created
  features: [{..}, {..}, ..]
}
```

### `draw.delete`

Fired when one or more features are deleted as the result of a user interaction. The following interactions will trigger this event:

[...]

PR #407, opened 2016-06-24
All new events

davidtheclark:
I have a first draft going of the new events discussed in #397.

@tmcw and @ghoshkaj, let's see how they work out in Studio.

Todo:

- [x] Ensure that programmatic changes (via API e.g. `Draw.deleteAll()`) do not fire events.
- [x] Update `Draw.changeMode()` docs.

Not yet tested:

- [x] Select and move or delete multiple vertices
- [x] Box-selection that catches no features
- [x] Enter key for all features
- [x] Escape key for all features

(If you think of anything else that needs testing please add it to the list.)


mcwhittemore:
> Rename draw.delete to draw.destroy

Why are we doing this? Are we having a problem with this name? The API names are `delete` and `deleteAll`, so changing this creates a disconnect between events and the api.

davidtheclark:
I noticed the API names after writing that. @tmcw was renaming things to destroy in Studio so I was looking into it. Don't have to do that.

API.md
```diff
@@ -205,80 +205,105 @@ The three drawing modes work identically. They do not take an options argument.
...
+  features: [{..}, {..}, ..]
 }
 ```
 
-### draw.modified
+### `draw.update`
+
+Fired when one or more features are updated as the result of a user interaction. The following interactions will trigger this event, with different `type` properties in the event data:
+
+- Finish moving one or more selected features in `simple_select` mode. The event will only fire when the movement is finished — i.e. when the user releases the mouse button or hits Enter. Type: `'move'`.
```
mcwhittemore: Should we start these with the type (ie: move) and then have the description of when it will happen?

mcwhittemore:
@davidtheclark This is looking really good. The notes in the docs about when to expect an event is great.

One of the pain points we had with `draw.changed` was that it fired too often. We tried to solve this problem with `draw.updated` by aggregating changes until a mouse up event happened. This had a few holes that need to be solved (not all changes are well define by mouse clicks) but it showed that aggregating events was a good first step. Lets find a way to fire this new `draw.update` event so that it can be aggregating without too much rework.

Currently the `mode_handler` has a provides a `fire` action to each mode. It does some magic with the event name that can be dropped as its not used anywhere anymore. Lets use this to fire events so that the mode_handler can provide aggregation logic as we move forward.

mcwhittemore:
Per delete vs destroy:

Cool. Lets not change names of things unless we find they current name creates problems.

davidtheclark:
> One of the pain points we had with draw.changed was that it fired too often

I do not see this problem with `draw.updated`, because it only fires at the end of interactions like dragging. I'm not sure what benefit additional aggregation logic would provide. Can you point out where the current events would need to be aggregated?

mcwhittemore:
Per delete vs destroy:

If we're ever going to make this change. Now is a good time, so lets do it.

davidtheclark:
Just pushed an update that I think clears up the question of how API invocations fire events.

davidtheclark:
With all the tests in place I'm feeling like these events are pretty solid. Have yet to demonstrate full integration with Studio, though. 

So @mcwhittemore it's ready for review (we can save any renaming for another PR), and do you think we should merge into master or leave it in a feature branch until we've demonstrated integration with Studio?

mcwhittemore:
@davidtheclark I'm looking now

test/interaction_events.test.js
```diff
@@ -0,0 +1,976 @@
...
+    map.fire('mousemove', makeMouseEvent(20, 20));
+    click(map, makeMouseEvent(25, 25));
+    afterNextRender(() => {
+      const pointId = firedWith(t, 'draw.create', {
+        features: [point25_25GeoJson]
+      }).features[0].id;
+
+      firedWith(t, 'draw.modechange', {
+        mode: 'simple_select',
+        options: { featureIds: [pointId] }
```
mcwhittemore: I thought we were dropping options?

davidtheclark: We could still drop them. I felt like I was able to clarify them a little just by ensuring it was always an object and had a `features` or `featuresId` array. Does that make enough sense you think to keep them?

davidtheclark: Also, I was just waiting to mess with them until I understood them better. I think I understand them now :)

src/modes/direct_select.js
```diff
@@ -67,9 +74,15 @@ module.exports = function(ctx, opts) {
         }
       });
       this.on('mouseup', () => true, function() {
+        if (dragging) {
+          ctx.map.fire(Constants.events.UPDATE, {
+            type: Constants.updateTypes.CHANGE_COORDINATES,
```
mcwhittemore: `type` is used by `mapbox-gl-js` to store the event fired type so when this event gets to the application its type is `draw.update`. Maybe action or reason will work?

davidtheclark: Sure, I like `action` fine.

src/modes/draw_line_string.js
```diff
@@ -20,7 +20,7 @@ module.exports = function(ctx) {
 
   function stopDrawingAndRemove() {
     ctx.events.changeMode(Constants.modes.SIMPLE_SELECT);
-    ctx.store.delete([line.id]);
+    ctx.store.delete([line.id], { silent: true });
```
mcwhittemore: Whats the difference between `silent` and `emit`?

src/modes/direct_select.js
```diff
@@ -107,6 +109,22 @@ module.exports = function(ctx, opts) {
...
+      }
+
+      selectedCoordPaths.sort().reverse().forEach(id => feature.removeCoordinate(id));
+      ctx.map.fire(Constants.events.UPDATE, {
+        type: Constants.updateTypes.CHANGE_COORDINATES,
+        features: ctx.store.getSelected().map(f => f.toGeoJSON())
+      });
+      selectedCoordPaths = [];
+      if (feature.isValid() === false) {
+        ctx.store.delete([featureId], { emit: options.emit });
```
mcwhittemore: Whats the difference between `silent` and `emit`?

davidtheclark: Any `emit` left in the code is a relic of earlier attempts and should be destroyed. Only `silent` should remain.

src/modes/draw_line_string.js
davidtheclark: `silent` means that it should not fire an event.

test/interaction_events.test.js
mcwhittemore: So, I think `draw.selectionchange` does what `opts` used to do. I think that makes sense. Even if we wanted to expose what Position has been selected I think we'd want to do it in `draw.selectionchange`

mcwhittemore:
@davidtheclark looks good to me. After you merge I'll rebase, fix my tests (write a new few) and merge. Than we can get this into Studio

test/interaction_events.test.js
davidtheclark: Ok. I will delete `options` from the event data, then, and we can re-add it if we ever see a need.

davidtheclark:
@mcwhittemore I had made some progress integrating the new events into Studio, but then just got stuck on the problem that `Draw.set()` is going to solve (every re-render lost the selection). So I think this will work out well.

PR #409, opened 2016-06-24
add a set function to the api

mcwhittemore:
Allows applications to hard reset Draw's internals without fully blowing away what is rendered on the map

### DIFF
--- a/API.md
+++ b/API.md
@@ -169,6 +169,24 @@ Draw
 
 ---
 
+### `.set(Object: featureCollection) -> [featureId, featureId]`
+
+This function takes a featureCollection and performs the required delete, create and update actions internally to make Draw represent this change. Effectively this is the same as `Draw.deleteAll()` followed by `Draw.add(featureCollection)` except that it doesn't effect performance as much.
+
+Example:
+
+```js
+var ids = Draw.set({type: 'FeatureCollection', features: [{
+  type: 'Feature',
+  properties: {},
+  id: 'example-id',
+  geometry: { type: 'Point', coordinates: [0, 0] }
+}]});
+// => ['example-id']
+```
+
+---
+
 ### `.trash() -> Draw`
 
 This envokes the current modes trash event. For the `simple_select` mode this deletes all active features. For the `direct_select` mode this deletes the active vertices. For the drawing modes, these cancel the current process.
--- a/API.md
+++ b/API.md
@@ -33,13 +33,13 @@ styles | Array | An array of style objects. By default draw provides a style for
 
 `mapboxgl.Draw()` returns an instance of `Draw` which has the following API for working with your data:
 
-###`.add(Object: GeoJSON) -> String || [String]`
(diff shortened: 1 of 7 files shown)


mcwhittemore:
@davidtheclark This is ready for your eyes.

API.md
```diff
@@ -169,6 +169,24 @@ Draw
...
+```js
+var ids = Draw.set({type: 'FeatureCollection', features: [{
+  type: 'Feature',
+  properties: {},
+  id: 'example-id',
+  geometry: { type: 'Point', coordinates: [0, 0] }
+}]});
+// => ['example-id']
+```
+
```
davidtheclark: Does it only accept FeatureCollections?

How is it intended to deal with partial features, e.g. could I do 

```js
Draw.set({ type: 'FeatureCollection', features: [{ 
  geometry: { 
    coordinates: [10, 10]
  }
}]});
```
... and it would know to update the coordinates of the first feature?

src/api.js
```diff
@@ -23,50 +23,52 @@ module.exports = function(ctx) {
...
-          type: 'Feature',
-          id: geojson.id,
-          properties: geojson.properties || {},
-          geometry: geojson
-        };
+    set: function(featureCollection) {
+      if (featureCollection.type === undefined || featureCollection.type !== 'FeatureCollection' || !Array.isArray(featureCollection.features)) {
+        throw new Error('Invalid FeatureCollection');
       }
+      var toDelete = JSON.parse(JSON.stringify(ctx.store.getAllIds()));
```
davidtheclark: Could clone this with a slice, right? `ctx.store.getAllIds().slice()` Maybe not ...

davidtheclark:
Seems to work great! My main concern (because of what I was seeing in Studio) was that it wouldn't mess up selections. And looks like it doesn't. I think the docs could use clarification (though I see the answers to those questions in the code and by trying it out). But we *could* save that for later. Merge when you feel like it.

API.md
davidtheclark: Answers based on reading the code below (which we should add to docs at some point — please correct me if they're wrong):

Yes, it only accepts FeatureCollections.

It compares features based on `id`. So any feature whose `id` matches a feature in the new collection will be replaced (completely, not merged) with the collection's feature.

src/api.js
mcwhittemore: I'm not sure which is faster, slice or concat. I'll find out and fix this.

API.md
mcwhittemore: > Does it only accept FeatureCollections?

It only accepts FeatureCollections

> How is it intended to deal with partial features, e.g. could I do

If the feature doesn't have an id it will create a new feature. If a feature doesn't have properties its invalid and an error will be thrown.

mcwhittemore:
I broke tests! Yes. I was wondering if my changes to add would cause a problem.

davidtheclark:
LGTM. Merging.

PR #414, opened 2016-06-25
Studio inspired fixes

davidtheclark:
- Add `Draw.getMode()` API function, so Studio could tell if it needed to update Draw's mode.
- Make `Draw.set()` re-render when it's done deleting.
- Prevent box selecting from happening because you shift+double-clicked to zoom out
- Exit `direct_select` mode if the selected feature has been deleted
- Add `Draw.setSelected()` to API as a preferable way to select items programmatically, instead of needing to change mode to `simple_select` and specifying features.
- Fix cursors so the right ones show and they change as the mouse moves.
- Fix this: create a feature, then before deselecting it try to drag it — it's like the browser tries to treat it as a file that you're dragging and dropping. Fixed by re-rendering the feature when you mousedown on it.

Might accumulate more ...

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -105,7 +105,7 @@ module.exports = function(ctx) {
         ctx.store.render();
       }
     },
-    changeMode: function(mode, modeOptions) {
+    changeMode: function(mode, modeOptions = {}) {
       // Avoid changing modes just to re-select what's already selected
       if (mode === Constants.modes.SIMPLE_SELECT && this.getMode() === Constants.modes.SIMPLE_SELECT) {
         if (stringSetsAreEqual((modeOptions.featureIds || []), ctx.store.getSelectedIds())) return;
--- a/src/api.js
+++ b/src/api.js
@@ -2,6 +2,7 @@ var isEqual = require('lodash.isequal');
 var normalize = require('geojson-normalize');
 var hat = require('hat');
 var featuresAt = require('./lib/features_at');
+var stringSetsAreEqual = require('./lib/string_sets_are_equal');
 var geojsonhint = require('geojsonhint');
 var Constants = require('./constants');
 
@@ -23,6 +24,15 @@ module.exports = function(ctx) {
     },
     getSelectedIds: function () {
       return ctx.store.getSelectedIds();
+    },
+    setSelected: function(idsToSelect) {
+      if (stringSetsAreEqual(idsToSelect, ctx.store.getSelectedIds())) return;
+      if (this.getMode() !== Constants.modes.SIMPLE_SELECT) {
+        ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: idsToSelect }, { silent: true });
+      } else {
+        ctx.store.setSelected(idsToSelect, { silent: true });
+        ctx.store.render();
+      }
     },
     set: function(featureCollection) {
       if (featureCollection.type === undefined || featureCollection.type !== 'FeatureCollection' || !Array.isArray(featureCollection.features)) {
@@ -98,14 +108,6 @@ module.exports = function(ctx) {
     },
     getMode: function() {
       return ctx.events.getMode();
-    },
-    setSelected: function(featureIds) {
-      if (this.getMode() === Constants.modes.SIMPLE_SELECT) {
-        ctx.store.setSelected(featureIds, { silent: true });
-      } else {
-        ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds }, { silent: true });
-      }
-      ctx.store.render();
     },
     trash: function() {
       ctx.events.trash({ silent: true });
--- a/src/api.js
+++ b/src/api.js
@@ -87,14 +87,25 @@ module.exports = function(ctx) {
     },
     delete: function(featureIds) {
       ctx.store.delete(featureIds, { silent: true });
-      ctx.store.render();
+      // If we were in direct select mode and our selected feature no longer exists
+      // (because it was deleted), we need to get out of that mode.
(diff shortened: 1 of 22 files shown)


davidtheclark:
Working on this in parallel with https://github.com/mapbox/studio/pull/7087

mcwhittemore:
> Make Draw.set() re-render when it's done.

This makes me thing that Draw.add or Draw.delete isn't rerendering as we'd like it to...

davidtheclark:
More:
- Prevent box selecting from happening because you shift+double-clicked to zoom out
- Exit `direct_select` mode if the selected feature has been deleted

davidtheclark:
> This makes me thing that Draw.add or Draw.delete isn't rerendering as we'd like it to...

@mcwhittemore True, I'll look into that ...

davidtheclark:
@mcwhittemore I think the issue is that at https://github.com/mapbox/mapbox-gl-draw/blob/studio-inspired-fixes/src/api.js#L42 `Draw.set()` is using the API's `add()` function, which does re-render when called, but at https://github.com/mapbox/mapbox-gl-draw/blob/studio-inspired-fixes/src/api.js#L52 it's using the store's `delete()` function, which does not re-render when called. I think it actually makes sense to use the store's methods in both places then save the rendering for the end of `Draw.set()`, right? To avoid rendering once for add and once for delete? Do you agree?

davidtheclark:
@mcwhittemore Nevermind ... I see that there's much more going on in API's `add` than in Store's, and `Draw.set()` needs that. I'll just use API's methods for both, for now.

mcwhittemore:
@davidtheclark I've tried to keep raw GeoJSON out of the store as when we use the store internally (in the `draw_*` modes) we want to add feature objects rather than geojson.

If the `Draw.delete()` took an array of feature ids we could use it in `Draw.set()` and get around this.

davidtheclark:
- Adds `Draw.setSelected()` to API as a preferable way to select items programmatically, instead of needing to change mode to `simple_select` and specifying features.

davidtheclark:
@mcwhittemore This is ready for review. I tried to put all the changes in a list in the first comment of this thread. Once we merge this into master, let's double-check docs, create issues for necessary tests, and try to get a release out.

src/api.js
```diff
@@ -82,8 +94,8 @@ module.exports = function(ctx) {
         features: ctx.store.getAll().map(feature => feature.toGeoJSON())
       };
     },
-    delete: function(id) {
-      ctx.store.delete([id], { silent: true });
+    delete: function(featureIds) {
```
mcwhittemore: This is a breaking change. We need to update the docs.

davidtheclark: It is actually not breaking: it will accept either a single ID, as before, or an array. But, yes, we need to clean up docs with these changes.

src/api.js
```diff
@@ -23,6 +25,15 @@ module.exports = function(ctx) {
     getSelectedIds: function () {
       return ctx.store.getSelectedIds();
     },
+    setSelected: function(idsToSelect) {
```
mcwhittemore: This is duplicate functionality to `Draw.changeMode()` that automatically changes the mode for you. Once we can do this without changing the mode I think its great, but until then lets just use what we have.

src/modes/simple_select.js
```diff
@@ -75,6 +79,15 @@ module.exports = function(ctx, options = {}) {
       // probably passed in from a `draw_*` mode
       if (ctx.store) ctx.store.setSelected(initiallySelectedFeatureIds);
 
+      // Any click should stop box selecting and dragging
+      this.on('click', () => true, function() {
+        dragging = false;
```
mcwhittemore: Should we move these to `cleanupBoxSelect` as well?

src/lib/mode_handler.js
```diff
@@ -37,15 +37,18 @@ var ModeHandler = function(mode, DrawContext) {
...
       var handle = handles[iHandle];
       if (handle.selector(event)) {
         handle.fn.call(ctx, event);
-        DrawContext.store.render();
-        break;
+        handlerCalled = true;
       }
     }
-    DrawContext.ui.updateMapClasses();
+    if (handlerCalled) {
```
mcwhittemore: I'm pretty sure this check is what caused the cursors to go away. Looks like you've fixed that problem by calling render a few times inside of the modes. Either way is good with me as long as the way we go is fastest :).

src/render.js
```diff
@@ -6,9 +6,15 @@ module.exports = function render() {
   if (!mapExists) return cleanup();
 
   var mode = store.ctx.events.currentModeName();
-  store.ctx.ui.queueMapClasses({
-    mode: mode
-  });
+
+  // If we were in direct select mode and our selected feature no longer exists
+  // (because it was deleted), we need to get out of that mode.
+  if (mode === Constants.modes.DIRECT_SELECT && !store.get(store.getSelectedIds()[0])) {
```
mcwhittemore: hmmm... this is ok for now, but I'd like to find a way to keep mode logic out of the renderer. Maybe there is a way to emit events only to the modes when its an application triggered event. This way modes could listen to the delete and selection change events and update themselves.

mcwhittemore:
@davidtheclark  Functionality wise this is looking good. Mouse cursors all seem right and moving between modes seems right. I haven't checked the events other than looking at the tests.

davidtheclark:
@mcwhittemore I think I addressed your concerns:

- Removed `Draw.setSelected`.
- Moved the direct-select-escape-hatch out of `render` (into the API functions `delete` and `deleteAll`).

Ready for another review or merge.

mcwhittemore:
@davidtheclark merged and published at 0.11.0.

PR #416, opened 2016-06-25
Replace SimpleSet with StringSet for performance

davidtheclark:
@mcwhittemore We talked a little late last week about trying to improve performance of SimpleSet for lots of data. That's what I tried to do here. Here are some benchmarks comparing the old array-based SimpleSet with the new object-based StringSet:

```
SimpleSet#constructor x 869,202 ops/sec ±4.06% (75 runs sampled)
StringSet#constructor x 8,254 ops/sec ±1.00% (88 runs sampled)
SimpleSet#add x 164,595 ops/sec ±1.95% (80 runs sampled)
StringSet#add x 56,942,286 ops/sec ±2.36% (85 runs sampled)
SimpleSet#delete x 172,353 ops/sec ±0.63% (82 runs sampled)
StringSet#delete x 12,259,264 ops/sec ±1.06% (82 runs sampled)
SimpleSet#has negative x 202,879 ops/sec ±1.20% (90 runs sampled)
StringSet#has negative x 60,145,554 ops/sec ±2.15% (82 runs sampled)
SimpleSet#has positive x 311,994 ops/sec ±0.54% (78 runs sampled)
StringSet#has positive x 73,808,838 ops/sec ±1.38% (88 runs sampled)
SimpleSet#values x 986,311 ops/sec ±0.82% (89 runs sampled)
StringSet#values x 40,677 ops/sec ±0.64% (82 runs sampled)
SimpleSet#clear x 70,565,649 ops/sec ±1.11% (85 runs sampled)
StringSet#clear x 56,828,172 ops/sec ±1.56% (85 runs sampled)
```

Performance is significantly better (or essentially the same) on everything except the `constructor` and `values`. Converting an array in the constructor to object keys is O(n), I believe; and the constructor in the benchmark was adding 1000 items. `StringSet.values()` uses `Object.keys()`, which I think is the fastest way available to get an array of an object's keys. But it's a lot slower than just returning an array you already have.

I also tried building an implementation that kept track of both an object and an array, so it could return that array more quickly for `StringSet.values()`. But that didn't improve the function's speed significantly, because in that implementation I ended up having to convert a possibly sparse array (if you've deleted items) to a dense array, which is O(n) just like `Object.keys()`. 

I also changed a couple of things in the Store to arrays because they did not gain from being Sets (did not ever perform delete or lookup).

What do you think?

(Also, this may be the first time I've ever tried to make statements about big-O notation, so please correct if I'm wrong.)


src/store.js
```diff
@@ -1,14 +1,14 @@
...
 
 var Store = module.exports = function(ctx) {
   this._features = {};
-  this._featureIds = new SimpleSet();
-  this._selectedFeatureIds = new SimpleSet();
-  this._changedFeatureIds = new SimpleSet();
-  this._deletedFeaturesToEmit = new SimpleSet();
+  this._featureIds = new StringSet();
+  this._selectedFeatureIds = new StringSet();
+  this._changedFeatureIds = [];
```
mcwhittemore: Why is this moving away from a set?

davidtheclark: Not 100% confident about this so interested in your opinion ... I changed that to an array because it was only ever being added to and then used as array. I figured that it might save a little time on each render to leave it as an array, as opposed to needing to convert from an the set's object to an array. However, this list is usually relatively small, so I don't know if that matters or not. What do you think?

mcwhittemore: I suspect that both ways are ok and that we wouldn't see much of a perf change because of the size. Might be good to keep this a Set just so all of the ids lists are sets.

mcwhittemore: I merged anyway as we can do this in one of the many other PRs coming down the pipe.

davidtheclark: > Might be good to keep this a Set just so all of the ids lists are sets.

Agreed. The consistency will be nice.

PR #424, opened 2016-06-28
fix Draw.add validation and add tests

### LINKED ISSUE #387: add()'s trying to fix geojson issues will only make them more complicated
https://github.com/mapbox/mapbox-gl-draw/blob/master/src/api.js#L27-L34

This should be more precise: let's check that the thing is actually a geometry, by testing for `type` being a valid geometry type. Otherwise we're making errors harder to parse by doing too much magic.

mcwhittemore:
Closes #387

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -55,8 +55,15 @@ module.exports = function(ctx) {
       var ids = featureCollection.features.map(feature => {
         feature.id = feature.id || hat();
 
-        if (ctx.store.get(feature.id) === undefined) {
+        if (feature.geometry === null) {
+          throw new Error('Invalid geometry: null');
+        }
+
+        if (ctx.store.get(feature.id) === undefined || ctx.store.get(feature.id).type !== feature.geometry.type) {
           var model = featureTypes[feature.geometry.type];
+          if (model === undefined) {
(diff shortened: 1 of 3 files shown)


mcwhittemore:
@davidtheclark for the review

src/api.js
```diff
@@ -55,8 +55,15 @@ module.exports = function(ctx) {
       var ids = featureCollection.features.map(feature => {
         feature.id = feature.id || hat();
 
-        if (ctx.store.get(feature.id) === undefined) {
+        if (feature.geometry === null) {
+          throw new Error('Invalid geometry: null');
+        }
+
+        if (ctx.store.get(feature.id) === undefined || ctx.store.get(feature.id).type !== feature.geometry.type) {
```
davidtheclark: > ctx.store.get(feature.id).type !== feature.geometry.type

What is that condition meant to catch?

davidtheclark:
Looks good. I had that one question —— We have a use-case for changing the geometry of an existing feature via `Draw.add()`? Is that what it's doing?

mcwhittemore:
@davidtheclark yep.

PR #427, opened 2016-06-28
Use xtend module instead of polyfill

davidtheclark:
[xtend](https://github.com/Raynos/xtend) is a super small module for doing what we need to do, and it is immutable by default, which we usually want. 

In addition to the unit tests I did a quick smoke test in `debug/` and nothing seemed to be broken.

Are you ok with me changing the debug port to 8080? Or anything else other than 9966, so I can run this server at the same time as Studio's ...

cc/ @mcwhittemore


package.json
```diff
@@ -18,7 +18,7 @@
     "coverage": "NODE_ENV=test nyc --reporter html npm run tape && opener coverage/index.html",
     "build": "NODE_ENV=production browserify index.js > dist/mapbox-gl-draw.js",
     "prepublish": "NODE_ENV=production browserify index.js | uglifyjs -c -m > dist/mapbox-gl-draw.js",
-    "start": "node server.js"
+    "start": "PORT=8080 node server.js"
```
mcwhittemore: We shouldn't override what PORT can do. If we need to change the default port, lets change in server.js

davidtheclark: You're right. After that is this good to merge?

PR #430, opened 2016-06-28
Consistify exiting draw modes

### LINKED ISSUE #399: Finishing a feature behaves unexpectedly
## Desired functionality 

- **Escape** should always destroy the feature
- **Enter** should clear the feature if at the last click the feature was not valid, other wise it should remove the currently being added leg and save the feature.
- **1-3** should behave like `enter`

## Current functionality

- Before one click
  - All features behave the same.
  -  clicking `escape`, `enter` or `1-3` results in no feature being added to the map
- After first click (LineString, Polygon)
  - LineString
    - `escape` removes the feature
    - `enter` removes the feature
    - `1-3` keeps the feature :bug:
  - Polygon
    - clicking `escape`, `enter` or `1-3` results in no feature being added to the map
- After second click
  - LineString
    - `escape` removes the feature
    - `enter` keeps one leg
    - `1-3` keeps the two legs :bug:
  - Polygon
    - `escape` removes the feature
    - `enter` removes the feature and throws an error :bug:
    - `1-3` keeps the feature :bug:
- After 3 or more clicks
  - LineString
    - `escape` removes the feature
    - `enter` removes most recently dragged out leg
    - `1-3` keeps all legs :bug:
  - Polygon
    - `escape` removes the feature
    - `enter` removes most recently dragged out leg
    - `1-3` keeps all legs :bug:

/cc @davidtheclark @samanpwbb

mcwhittemore:
Closes #399

@davidtheclark for the review

### DIFF
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -35,21 +35,11 @@ module.exports = function(ctx) {
     if (currentVertexPosition > 0 &&
       (isEventAtCoordinates(e, line.coordinates[0]) || isEventAtCoordinates(e, line.coordinates[currentVertexPosition - 1]))
     ) {
-      return finish();
+      return ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [line.id] });
     }
 
     line.updateCoordinate(currentVertexPosition, e.lngLat.lng, e.lngLat.lat);
     currentVertexPosition++;
-  }
-
-  function finish() {
-    if (!line.isValid()) return stopDrawingAndRemove();
-    line.removeCoordinate(`${currentVertexPosition}`);
-    currentVertexPosition--;
-    ctx.map.fire(Constants.events.CREATE, {
-      features: [line.toGeoJSON()]
-    });
-    ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [line.id] });
   }
 
   return {
@@ -61,16 +51,23 @@ module.exports = function(ctx) {
       this.on('mousemove', CommonSelectors.true, handleMouseMove);
       this.on('click', CommonSelectors.true, handleClick);
       this.on('keyup', CommonSelectors.isEscapeKey, stopDrawingAndRemove);
-      this.on('keyup', CommonSelectors.isEnterKey, finish);
+      this.on('keyup', CommonSelectors.isEnterKey, () => {
+        ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [line.id] });
+      });
     },
 
     stop() {
       doubleClickZoom.enable(ctx);
       ctx.ui.setActiveButton();
 
-      // If it's invalid, just destroy the thing
-      if (!line.isValid()) {
-        ctx.store.delete([line.id], { silent: true });
+      line.removeCoordinate(`${currentVertexPosition}`);
+      if (line.isValid() && ctx.store.get(line.id) !== undefined) {
+        ctx.map.fire(Constants.events.CREATE, {
+          features: [line.toGeoJSON()]
+        });
+      }
+      else if (ctx.store.get(line.id) !== undefined) {
+        stopDrawingAndRemove();
       }
     },
 
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -58,13 +58,14 @@ module.exports = function(ctx) {
       doubleClickZoom.enable(ctx);
       ctx.ui.setActiveButton();
 
+      if (ctx.store.get(line.id) === undefined) return;
       line.removeCoordinate(`${currentVertexPosition}`);
-      if (line.isValid() && ctx.store.get(line.id) !== undefined) {
+      if (line.isValid()) {
         ctx.map.fire(Constants.events.CREATE, {
           features: [line.toGeoJSON()]
         });
       }
-      else if (ctx.store.get(line.id) !== undefined) {
+      else {
         ctx.store.delete([line.id], { silent: true });
         ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, {}, { silent: true });
       }
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -19,11 +19,6 @@ module.exports = function(ctx) {
 
   ctx.store.add(line);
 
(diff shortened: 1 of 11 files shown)


src/modes/draw_line_string.js
```diff
@@ -60,17 +45,28 @@ module.exports = function(ctx) {
...
+      });
     },
 
     stop() {
       doubleClickZoom.enable(ctx);
       ctx.ui.setActiveButton();
 
-      // If it's invalid, just destroy the thing
-      if (!line.isValid()) {
+      line.removeCoordinate(`${currentVertexPosition}`);
```
davidtheclark: Are we able to remove the coordinate if the line is not valid?

src/modes/draw_line_string.js
```diff
@@ -60,17 +45,28 @@ module.exports = function(ctx) {
...
+      if (line.isValid() && ctx.store.get(line.id) !== undefined) {
```
davidtheclark: When can you arrive here with an undefined feature?

src/modes/draw_line_string.js
```diff
@@ -60,17 +45,28 @@ module.exports = function(ctx) {
...
+        ctx.map.fire(Constants.events.CREATE, {
+          features: [line.toGeoJSON()]
+        });
+      }
+      else if (ctx.store.get(line.id) !== undefined) {
         ctx.store.delete([line.id], { silent: true });
+        ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, {}, { silent: true });
```
davidtheclark: I think the repetition of these two lines in 3 places justifies something like the `finish` function, right?

src/modes/draw_polygon.js
```diff
@@ -58,29 +43,41 @@ module.exports = function(ctx) {
...
+      });
     },
 
     stop: function() {
       doubleClickZoom.enable(ctx);
       ctx.ui.setActiveButton();
 
-      // If it's invalid, just destroy the thing
-      if (!polygon.isValid()) {
+      polygon.removeCoordinate(`0.${currentVertexPosition}`);
```
davidtheclark: I think I ran into a bug where this failed when the polygon was not valid.

src/modes/draw_polygon.js
```diff
@@ -58,29 +43,41 @@ module.exports = function(ctx) {
...
+      if (polygon.isValid() && ctx.store.get(polygon.id) !== undefined) {
+        ctx.map.fire(Constants.events.CREATE, {
+          features: [polygon.toGeoJSON()]
+        });
+      }
+      else if (ctx.store.get(polygon.id) !== undefined) {
         ctx.store.delete([polygon.id], { silent: true });
+        ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, {}, { silent: true });
       }
```
davidtheclark: These conditionals could be simplified a little by early returning if the feature is undefined:

```js
if (ctx.store.get(polygon.id) === undefied) return;
if (polygon.isValid()) {
 ..
} else {
 ..
}
```

test/draw_line_string.test.js
```diff
@@ -295,52 +276,51 @@ test('draw_line_string interaction', t => {
...
-      click(map, makeMouseEvent(1, 1));
+      mouseClick(map, makeMouseEvent(1, 1));
+      map.fire('mousemove', makeMouseEvent(16, 16));
       st.equal(Draw.getAll().features.length, 1, 'no longer drawing');
 
       st.end();
     });
 
-    t.test('start draw_line_string mode then exit before a click', st => {
+    t.test('start draw_line_string mode then change_mode before a click', st => {
```
davidtheclark: change mode without underscore, right? underscore makes it seem like it's a mode

test/utils/key_events.js
```diff
@@ -0,0 +1,21 @@
...
+export const startPointEvent = createSyntheticEvent('keydown', {
+  keyCode: 49
+});
+
+export const startLineStringEvent = createSyntheticEvent('keydown', {
+  keyCode: 50
+});
+
+export const startPolygonEvent = createSyntheticEvent('keydown', {
+  keyCode: 51
```
davidtheclark: Are the 1-3 keyboard interactions documented? I don't see them.

src/modes/draw_line_string.js
mcwhittemore: yep. tests cover both cases.

src/modes/draw_line_string.js
mcwhittemore: Line is a local scoped var so deleting the line from the store doesn't kill this. If we hit escape and get here the line could be valid be deleted.

src/modes/draw_polygon.js
mcwhittemore: Can you explain this a bit more. I'm not seeing it.

davidtheclark: If I remember right, I think I maybe have hit this error when I entered polygon draw mode then double clicked on the map. May have just been an error I saw in the console, not something that broke the app.

src/modes/draw_line_string.js
davidtheclark: Makes sense. A comment in the code would be nice to explain that context.

test/utils/key_events.js
mcwhittemore: Hmmm... these are part of the point, line, and polygon controls. We should fill out those docs.

davidtheclark:
These are great tests and code cleanups! I had some questions that you see.

test/utils/key_events.js
davidtheclark: If you want to just make another issue for that, that'd be swell.

mcwhittemore: - [x] done

src/modes/draw_polygon.js
mcwhittemore: hmmm... I'm not seeing this locally.
check

davidtheclark: Ok, maybe something else has taken care of that. If it's not happening, let's forget about it.

mcwhittemore:
@davidtheclark tests failed once because of a flap... looks like maybe something with `draw.render` events not being fired in a test where we expect it to and than never ending. I'm not sure if its something about the changes here, just want to make sure we have our eyes out for it so we can crush it when/if it flaps again.

PR #434, opened 2016-06-28
Cleanup and extend API tests

davidtheclark:
@mcwhittemore:

Almost all of the changes in `api.js` are just indent changing, because I created the `api` object before the methods so it was easy to return without worrying about `this` context.

I have one test set failing because of `map.queryRenderedFeatures()` ...

### DIFF
--- a/test/api.test.js
+++ b/test/api.test.js
@@ -62,12 +62,12 @@ test('Draw.set', t => {
     type: 'FeatureCollection',
     features: [point, line, polygon]
   };
-  const returnValue = Draw.set(collection);
-  t.equal(returnValue.length, 3,
+  const drawInstance = Draw.set(collection);
+  t.equal(drawInstance.length, 3,
     'return value is correct length');
-  const pointId = returnValue[0];
-  const lineId = returnValue[1];
-  const polygonId = returnValue[2];
+  const pointId = drawInstance[0];
+  const lineId = drawInstance[1];
+  const polygonId = drawInstance[2];
   t.equal(Draw.get(pointId).geometry.type, 'Point',
     'point id returned');
   t.equal(Draw.get(lineId).geometry.type, 'LineString',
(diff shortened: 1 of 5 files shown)


davidtheclark:
@mcwhittemore ready for review at last.

test/api.test.js
```diff
@@ -1,169 +1,364 @@
...
 
-    Draw.changeMode('draw_'+Draw.types.POINT);
-    var featureCollection = Draw.getAll();
-    t.equals(featureCollection.features[0].geometry.type, 'Point', 'Point');
+  // First set it to one collection
+  const collection = {
+    type: 'FeatureCollection',
+    features: [point, line, polygon]
+  };
+  const returnValue = Draw.set(collection);
```
mcwhittemore: Lets use a better var name here.

test/api.test.js
```diff
@@ -1,169 +1,364 @@
...
-    t.equals(featureCollection.features[0].geometry.type, 'LineString', 'Line');
-    Draw.deleteAll();
+  // Then set to another
+  addSpy.reset();
+  deleteSpy.reset();
+  const nextCollection = {
+    type: 'FeatureCollection',
+    features: [polygon]
+  };
+  const nextReturnValue = Draw.set(nextCollection);
```
mcwhittemore: Here as well.

test/api.test.js
```diff
@@ -1,169 +1,364 @@
...
+  addSpy.reset();
+  deleteSpy.reset();
+  const newLine = getGeoJSON('line');
+  const overlappingPolygon = getGeoJSON('polygon');
+  overlappingPolygon.id = nextPolygonId;
+  const overlappingCollection = {
+    type: 'FeatureCollection',
+    features: [newLine, overlappingPolygon]
+  };
+  const overlappingReturnValue = Draw.set(overlappingCollection);
```
mcwhittemore: Why `ReturnValue`? This seems like a var name you wouldn't normally use?

davidtheclark: I guess I was just being generic because we're testing the return value. I can change all "returnValue" to "drawInstance", so it matches the expected result.

mcwhittemore: yea, it started to make more sense as I read the tests, but when reading one test its pretty confusing about why. Lets name them based on what we expect them to return so help future us.

mcwhittemore:
@davidtheclark other than using something other than `returnValue` this looks great. Once that's updated lets merge.

PR #435, opened 2016-06-29
[WIP] Add static mode

drewbo:
Ref #109 cc: @mcwhittemore 

Happy to make any adjustments to make this more resilient, can add a few tests too


mcwhittemore:
@drewbo thanks for this. I know turning of interactivity is something a bunch of people are interested in so its good to see movement here.

Here are the things I think are left to do.

- [ ] update the `Draw.changeMode()` doc to describe this mode. (note there is no way to get to it or out of it except via `Draw.changeMode()`)
- [ ] update the styling docs to mention this mode
- [ ] update the default style to handle this mode
- [ ] write a few tests that confirm this mode doesn't do the same thing as `simple_select`.
- [ ] fix linting errors

src/modes/static.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function(ctx) {
+  return {
+    stop: function() {
+      ctx.map.doubleClickZoom.enable();
```
davidtheclark: This can use the `doubleClickZoom` lib function: https://github.com/mapbox/mapbox-gl-draw/blob/master/src/lib/double_click_zoom.js

davidtheclark: Actually, though, I'm confused as to why this needs to enable the zoom. Don't the other modes that disable zoom clean themselves up?

drewbo: Yup, I think this is mostly me being sloppy when I was first adding it for myself. I'll remove.

drewbo:
@mcwhittemore made the changes above with two remaining items to hit that I could use some assistance on:
 - `options.test.js` is now failing because `fixtures/style_with_sources.json` needs to be updated to reflect the new base style. What's the easiest way to update the fixture?
 - I had to add `map.dragPan.disable.restore();` before wrapping that method in `static.test.js` because it is already wrapped from `simple_select.test.js`. The new test runs fine on its own without that line. Should the spy be "unwrapped" in `simple_select.test.js` instead?

mcwhittemore:
> What's the easiest way to update the fixture?

Oh man. Not sure. Maybe you can console log it out.

> Should the spy be "unwrapped" in simple_select.test.js instead?

Yea, that sounds good to me.

drewbo:
@mcwhittemore updated the fixture. I left the other issue as is because I couldn't reliably `.restore` the method in `simple_select.test.js` and still have everything work. Not totally positive what the issue is but the tests do pass now

mcwhittemore:
@drewbo this is looking good. The only problem I see is that the cursor still changes to 👆 hand when I hover over the feature in static mode. Lets fix this up (add mode classes to the cursors css) and then rebase and we should be good.

Thanks!

drewbo:
@mcwhittemore will do. Do we still need to rebase if we can squash the commits with GH?

mcwhittemore:
@drewbo um, as long as the `out-of-date` warning goes away I'm happy.

drewbo:
Not positive what the `out-of-date` warning is but I added the styling

mcwhittemore:
@drewbo I'm talking about the warning git hub is displaying above the comment box.

image

drewbo:
Ah, got it. I don't see those but I merged master on.

mcwhittemore:
@drewbo merged and published (0.11.2). We're keeping the 0.10.x line as the main line in npm now so make sure you install 0.11.2 specifically.

PR #436, opened 2016-06-29
Mode handler tests; closes #370

### LINKED ISSUE #370: Add unit tests for lib files

davidtheclark:
@mcwhittemore for review.


mcwhittemore:
This si great.

PR #438, opened 2016-06-29
Render after changing selection

davidtheclark:
I found this bug while working in Studio. We need to make sure that if the selection is changed via `Draw.changeMode()`, there is a re-rendering.

@mcwhittemore for review.


mcwhittemore:
Looks good to me. Sometime (not too soon) we should remove these checks and get moving between modes to be smarter so we can avoid this kind of bug.

PR #439, opened 2016-06-30
Change server port for Studio compat

davidtheclark:
@mcwhittemore for review.

We talked about this the other day ... I just want to be able to run Draw and Studio servers at the same time easily. Open to any port number if you don't like that one.


mcwhittemore:
my api key is in local storage at 9967. Is that good with you?

davidtheclark:
Sounds good. Modified the PR.

PR #441, opened 2016-07-01
No useless render on api.add

davidtheclark:
This little conditional prevents `Draw.add` from calling `render` if it has nothing to add. 

I think it should do no harm to add ... and for some reason it seems to fix a bug I'm seeing in Studio. The trouble is, I don't know *why*. 

The bug is that sometimes when Studio uses `Draw.set`, a feature is not deleted, *or is only partially deleted* (i.e. you can see fragments of it on the map). As far as we can tell, this only happens when you are *editing* (not initially creating) an *empty dataset*: create a polygon, deselect it, select it, delete it — and the whole or a fragment sometimes remains.

The fragmentation, when it happens, looks like it's along tile boundaries.

The fact that this conditional *seems* to fix that apparent bug suggests that the root cause has something to do with either

- Draw's `throttle` function not working right, or
- Mapbox GL JS having trouble with Draw using `setData()` multiple times in rapid succession

Either way, it seems bananas that this only happens on empty datasets ... 

Any ideas, @mcwhittemore??

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -78,7 +78,10 @@ module.exports = function(ctx) {
       return feature.id;
     });
 
-    ctx.store.render();
+    if (ids.length) {
+      ctx.store.render();
+    }
+
     return ids;
   };
 


mcwhittemore:
Hmmm... 

I feel like something else must be going on here. I've seen some weird things happen with trashing the render call but Draw should only be rendering once every 16ms and we push this with dragging a lot so its more complex than calling setData.

The fact that we're seeing this with deleted things is also really interesting.

I'll spend sometime tomorrow combing through to see what could be happening.

mcwhittemore:
Ok. Looking deeper I think there is a problem that happens when we tell Draw.set to render an empty feature collection. It can result in the gl showing the feature that was being deleted already. Still trying to understand.

mcwhittemore:
@davidtheclark ok, so the problem here is that gl-js can't take this quick of a render change. I'm pretty sure. We see it on delete because its clear that the tiles are out of date. This only happens if you quickly render.

Part of me thinks that throttle should be stopping this from happening... but I'm going to try bring in lodash.throttle to see if that fixes it.

PR #444, opened 2016-07-01
Prevent box-selecting outside canvas; fixes #419

### LINKED ISSUE #419: Box selecting outside browser viewport breaks
If I shift click to box select and then move the cursor outside the browser viewport, the box is still rendered but deactivated. I return the cursor to the viewport and try to continue the box selection, but it is essentially broken.

jun-27-2016 13-14-22

davidtheclark:
Ref #419.

Adding the `mouseout` listener is a stopgap measure until mapbox-gl-js has a release including mapbox/mapbox-gl-js#2777. Once that's in, we can use the Map's `mouseout` event.

The reason I made `events.initialize` was so that the mode handler is only created after the map is available.

@mcwhittemore, what do you think of this solution? If you agree with it, I'll modify the tests so they pass. If you have another idea, let me know.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -20,7 +20,23 @@ module.exports = function(ctx) {
 
   var events = {};
   var currentModeName = Constants.modes.SIMPLE_SELECT;
-  var currentMode = ModeHandler(modes.simple_select(ctx), ctx);
+  var currentMode;
+
+  function initialize() {
+    // Add event listeners
+    ctx.map.on('mousemove', events.mousemove);
+
+    ctx.map.on('mousedown', events.mousedown);
+    ctx.map.on('mouseup', events.mouseup);
+
+    if (ctx.options.keybindings) {
+      ctx.container.addEventListener('keydown', events.keydown);
+      ctx.container.addEventListener('keyup', events.keyup);
+    }
+
+    // Start us in simple_select mode
+    currentMode = ModeHandler(modes.simple_select(ctx), ctx);
+  }
 
   events.drag = function(event) {
     if (isClick(mouseDownInfo, {
@@ -124,6 +140,7 @@ module.exports = function(ctx) {
   }
 
   var api = {
+    initialize,
     changeMode,
     currentModeName: function() {
       return currentModeName;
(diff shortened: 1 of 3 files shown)


src/events.js
```diff
@@ -20,7 +20,23 @@ module.exports = function(ctx) {
...
+    ctx.map.on('mousedown', events.mousedown);
+    ctx.map.on('mouseup', events.mouseup);
+
+    if (ctx.options.keybindings) {
+      ctx.container.addEventListener('keydown', events.keydown);
+      ctx.container.addEventListener('keyup', events.keyup);
+    }
+
+    // Start us in simple_select mode
+    currentMode = ModeHandler(modes.simple_select(ctx), ctx);
```
mcwhittemore: Whats the goal of this change?

davidtheclark: The goal was to ensure that `ctx.map` is available in `simple_select`'s `start` method, so we can get its container and attach a listener to it. The way things were before, `simple_select` was first initialized before the map was ready.

src/modes/simple_select.js
```diff
@@ -10,6 +10,7 @@ module.exports = function(ctx, options = {}) {
   var startPos = null;
   var featureCoords = null;
   var startCoordinates = null;
+  var mapCanvas = ctx.map.getCanvasContainer();
```
mcwhittemore: I'd prefer to not have refs to the map as part of the mode. This is the goal of the events.js. Adding a new handler to `mouseout` like we have for `keyup` and `keydown` would be better.

src/modes/simple_select.js
```diff
@@ -82,13 +91,15 @@ module.exports = function(ctx, options = {}) {
         return ctx.store.get(id) !== undefined;
       }));
 
-      // Any mouseup should stop box selecting and dragging
-      this.on('mouseup', () => true, function() {
+      // Attach a listener to mouseout, pending the mapbox-gl-js release including
+      // https://github.com/mapbox/mapbox-gl-js/pull/2777
+      mapCanvas.addEventListener('mouseout', handleMouseout);
```
mcwhittemore: This is a good idea, lets use `this.on('mouseout', () => true, function() {});` rather than a canvas map handler.

src/modes/simple_select.js
```diff
@@ -69,8 +73,13 @@ module.exports = function(ctx, options = {}) {
     featureCoords = featureIds.map(id => ctx.store.get(id).getCoordinates());
   };
 
+  var handleMouseout = function() {
+    cleanupBoxSelect();
+  };
+
   return {
     stop: function() {
+      mapCanvas.removeEventListener('mouseout', handleMouseout);
```
mcwhittemore: We should be able to drop this once `mouseout` is handled by events.js

src/events.js
mcwhittemore: I'm pretty sure this is happening as the mode was connecting to the DOM. modes shouldn't connect this way so I think we can go back to how it was before.

I only want to go back to how it was before because I'm not 100% on what this change might mean. Also, I think that have an initialize function in what is effectively an initialize function is a bit confusing.

davidtheclark:
I am going to use these ideas to start a new PR.

PR #446, opened 2016-07-01
Batch render calls invoked by Draw.set

mcwhittemore:
`Draw.set()` invokes render twice. Once for `Draw.add()` and once for `Draw.delete()`. Since we use `throttle` to limit the number of times we processes all of the renders this means that part of the job is sent to gl-js, then 16ms happens, and than another part of the job is sent. This can lead to some weird rendering results (due to gl-js not ensuring setData calls are rendered in the right order?). This PR solves this problem by batching the two renders into one.

This could also be solved by moving from throttle to debounce, but debounced rendering feels laggy.

Replaces #441

@davidtheclark for the review

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "hat": "0.0.3",
     "lodash.isequal": "^4.2.0",
     "point-geometry": "0.0.0",
-    "sinon": "^1.17.4",
     "turf-area": "^1.1.1",
     "xtend": "^4.0.1"
   }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -72,7 +72,6 @@
     "hat": "0.0.3",
     "lodash.isequal": "^4.2.0",
     "point-geometry": "0.0.0",
-    "sinon": "^1.17.4",
```
davidtheclark: Good catch!

davidtheclark:
Looks great! Merge away.

Issue #440, opened 2016-06-30
Box selecting starts when you shift+double-click to zoom out, does not stop


davidtheclark:
If you shift+double-click to zoom out, then click somewhere else on the map, you'll see that box select has started with the shift+double-click and now you have to finish it.

mcwhittemore:
It looks like anytime you use shift in simple_select your next interaction is box select.

I'm seeing this right now when I select two features (shift down) and than drag the map (shift up)

davidtheclark:
Closed by #447.

PR #447, opened 2016-07-01
Shift clicking

davidtheclark:
Ref #440.

This was one of those cases where to debug the problem I just had to clarify all the logic, so I understood it better and knew there weren't false moves.

So I refactored simple_select and I think it's more clear and more precise now. I believe clarifying process fixed our shift-click problems. 

The UI works as before, from what I can tell, and tests are passing (I removed two that seemed unnecessary, because UI confirmed that double-click zoom was behaving normally).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mapbox-gl-draw",
-  "version": "0.11.4",
+  "version": "0.11.5",
   "description": "A drawing component for Mapbox GL JS",
   "homepage": "https://github.com/mapbox/mapbox-gl-draw",
   "author": "mapbox",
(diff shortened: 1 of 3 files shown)


davidtheclark:
@mcwhittemore for review.

mcwhittemore:
bad-select

mcwhittemore:
@davidtheclark this looks pretty good. I found one bug.

davidtheclark:
@mcwhittemore I think I reproduced that and fixed it. The problem was that you could select multiple features then shift-click to drag move and box select at the same time. Now it will just drag move (not box select).

PR #449, opened 2016-07-02
Add mouseout and use it to cancel box select; fixes #419

### LINKED ISSUE #419: Box selecting outside browser viewport breaks

davidtheclark:
Re-take on PR #444 to close bug #419.

This time I used `events.js` to add the DOM listener, and added some appropriate tests.

@mcwhittemore, ready for review.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -14,10 +14,7 @@ var modes = {
 
 module.exports = function(ctx) {
 
-  var mouseDownInfo = {
-    isDown: false
-  };
-
+  var mouseDownInfo = {};
   var events = {};
   var currentModeName = Constants.modes.SIMPLE_SELECT;
   var currentMode = ModeHandler(modes.simple_select(ctx), ctx);
@@ -36,7 +33,7 @@ module.exports = function(ctx) {
   };
 
   events.mousemove = function(event) {
-    if (mouseDownInfo.isDown) {
+    if (event.originalEvent.which === 1) {
       return events.drag(event);
     }
     var target = getFeaturesAndSetCursor(event, ctx);
@@ -46,7 +43,6 @@ module.exports = function(ctx) {
 
   events.mousedown = function(event) {
     mouseDownInfo = {
-      isDown: true,
       time: new Date().getTime(),
       point: event.point
     };
@@ -56,7 +52,6 @@ module.exports = function(ctx) {
   };
 
   events.mouseup = function(event) {
-    mouseDownInfo.isDown = false;
     var target = getFeaturesAndSetCursor(event, ctx);
     event.featureTarget = target;
 
--- a/src/lib/move_feature.js
+++ b/src/lib/move_feature.js
@@ -1,0 +1,25 @@
+const Constants = require('../constants');
+
+module.exports = function(feature, lngDelta, latDelta) {
+  const featureCoordinates = feature.getCoordinates();
+
+  const moveCoordinate = (coord) => [coord[0] + lngDelta, coord[1] + latDelta];
+  const moveRing = (ring) => ring.map(coord => moveCoordinate(coord));
+  const moveMultiPolygon = (multi) => multi.map(ring => moveRing(ring));
+
+  if (feature.type === Constants.geojsonTypes.POINT) {
+    return feature.incomingCoords(moveCoordinate(featureCoordinates));
+  }
+
+  if (feature.type === Constants.geojsonTypes.LINE_STRING || feature.type === Constants.geojsonTypes.MULTI_POINT) {
+    return feature.incomingCoords(featureCoordinates.map(moveCoordinate));
+  }
+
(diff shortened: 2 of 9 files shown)


mcwhittemore:
Rebasing

mcwhittemore:
@davidtheclark this is looking good. There is a bug with mouseout when dragging a vertex as well.

mouse out drag

davidtheclark:
Additionally, weird stuff happens if you try to drag-move a feature east or west beyond the canvas.

davidtheclark:
This might be an issue with mapbox-gl-js. It seems like if I mousedown then drag off the canvas I can dragPan even if I've disabled drag pan and it doesn't work *on* the canvas.

davidtheclark:
@mcwhittemore With the latest commit, feature drag-moving should stop when you leave the canvas, without any weird dragPanning taking its place.

If you get a minute and can review this from a UX perspective, that'd be great. If we like the interaction, I'll modify the tests for them.

Another related bug that may need to be addressed as part of this: I shouldn't be able to warp features by  drag-moving such that part of them moves past a pole, north or south. Probably we need to say that whenever the northern-most vertex of any selected feature hits the north pole (or reverse for south), the movement stops?

davidtheclark:
@mcwhittemore Per the scrum conversation, I'll save the dragging-past-poles bug for another PR. And I guess I didn't break any tests with this one, so it's ready for review & maybe merge.

mcwhittemore:
@davidtheclark I'm seeing a bug with dragging a point off the screen.

drag_out

Also, should it stop drag in simple_select or just hold off on it until the mouse comes back in...

davidtheclark:
@mcwhittemore That direct_select problem was a silly last minute mistake I made before pushing. Should be fixed now.

jul-05-2016 11-53-04

And yes, I guess it would be nice if the moving just held off until you returned to the canvas. I'll work on that.

davidtheclark:
So, turns out we may not need the `mouseout` functionality after all. With the latest commit. You can drag off the canvas and the movement will stop, then it will resume when you bring the mouse back on the canvas. Same goes for box select. 

Seems to work pretty well. The only thing about this that I think is a little funny is that you can release the mouse while off-canvas and when you go back on-canvas you're continuing the interaction as though you we were still dragging. I'm worried we may have to choose between this weirdness, or the weirdness (from before the latest commit) of fully cancelling your drag event when you leave the canvas even if you keep holding the mouse key down.

davidtheclark:
Ah, good thinking. I guess if the buttons aren't down on a mousemove we could cancel the action.

davidtheclark:
Ok, @mcwhittemore, I think it's on track now. I learned that the internal `drag` and `mousemove` events are mutually exclusive based on the mouse's buttons, so I added a couple of lines to cancel these extended interactions if there's a mousemove event that is not a drag.

mcwhittemore:
@davidtheclark sadly the difference between mousemove and drag in Draw is if the mouseup event happened and since mouseup events don't happen when the mouse is off this doesn't work.

That said, I like the idea here. Lets have the event.js set `mouseDownInfo.isDown = false` on a mouseout event.

mcwhittemore:
Also, it looks like there is a `which` attribute on MouseEvents. But MDN tells you to not use it so I'd say we shouldn't use it.

davidtheclark:
> sadly the difference between mousemove and drag in Draw is if the mouseup event happened

Could we not at this line https://github.com/mapbox/mapbox-gl-draw/blob/master/src/events.js#L39 use `event.originalEvent.buttons !== 1` instead of the `mouseDownInfo`?

mcwhittemore:
Looks like it doesn't work on Safari...

image

davidtheclark:
Damn this is has proven to be quite a pain. One more commit to fix it all ...

davidtheclark:
Agh, that puts us back in our original situation where it doesn't matter whether you hold the mouse down or not when you go off canvas ...

mcwhittemore:
@davidtheclark I'm good with this commit https://github.com/mapbox/mapbox-gl-draw/pull/449/commits/aefb1fa366ac0521c67368df9893f08c2f6a75bd

davidtheclark:
Ok, @mcwhittemore, here it is. Summary of the changes:

- Mouse events now determine whether we have `drag` or `mousemove` based on whether `event.which === 1` (that's drag). This is so that if a user drags off the canvas, mouseups off canvas, then moves the mouse back on canvas, the drag does not resume.
- `moveFeatures()` is its own external function. This happened during an experiment but makes sense, I think, so I kept it; and it will make even more sense once the next adjustment comes in (which will prevent you from moving parts of features past poles).
- Fixed the bug in both `simple_select` and `direct_select` mode where if you dragged off-canvas the map's dragPan kicked in right away, causing weird jumps.
- Updated and added tests.

mcwhittemore:
This looks good. Rebasing. Once the tests pass, I'll merge.

PR #450, opened 2016-07-02
Constants everywhere; closes #415

### LINKED ISSUE #415: Expand and normalize use of constants
I think that the events and modes now use constants pretty consistently. Worth double-checking.

Other constant-worthy strings should also use constants consistently. For example, class names can be stored in constants so it's easy to view them all in one place (right now there are some which are sometimes used, but we should standardize that).

davidtheclark:
Ref #415.

@mcwhittemore for review.

If you don't think this (or some of this) is worth constantizing, let me know! I'd like to know more about why people like or don't like constants, and in which places.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,12 +1,21 @@
 module.exports = {
-  CONTROL_BUTTON_CLASS: 'mapbox-gl-draw_ctrl-draw-btn',
-  CONTROL_BUTTON_LINE_CLASS: 'mapbox-gl-draw_line',
-  CONTROL_BUTTON_POLYGON_CLASS: 'mapbox-gl-draw_polygon',
-  CONTROL_BUTTON_POINT_CLASS: 'mapbox-gl-draw_point',
-  CONTROL_BUTTON_TRASH_CLASS: 'mapbox-gl-draw_trash',
-  CONTROL_GROUP_CLASS: 'mapboxgl-ctrl-group',
-  ATTRIBUTION_CLASS: 'mapboxgl-ctrl-attrib',
-  ACTIVE_BUTTON_CLASS: 'active',
+  classes: {
+    CONTROL_BASE: 'mapboxgl-ctrl',
+    CONTROL_PREFIX: 'mapboxgl-ctrl-',
+    CONTROL_BUTTON: 'mapbox-gl-draw_ctrl-draw-btn',
+    CONTROL_BUTTON_LINE: 'mapbox-gl-draw_line',
+    CONTROL_BUTTON_POLYGON: 'mapbox-gl-draw_polygon',
+    CONTROL_BUTTON_POINT: 'mapbox-gl-draw_point',
+    CONTROL_BUTTON_TRASH: 'mapbox-gl-draw_trash',
+    CONTROL_GROUP: 'mapboxgl-ctrl-group',
+    ATTRIBUTION: 'mapboxgl-ctrl-attrib',
+    ACTIVE_BUTTON: 'active',
+    BOX_SELECT: 'mapbox-gl-draw_boxselect'
+  },
+  sources: {
+    HOT: 'mapbox-gl-draw-hot',
+    COLD: 'mapbox-gl-draw-cold'
+  },
   cursors: {
     ADD: 'add',
     MOVE: 'move',
@@ -18,6 +27,17 @@ module.exports = {
     POLYGON: 'polygon',
     LINE: 'line_string',
     POINT: 'point'
+  },
+  geojsonTypes: {
+    FEATURE: 'Feature',
+    POLYGON: 'Polygon',
+    LINE_STRING: 'LineString',
+    POINT: 'Point',
+    FEATURE_COLLECTION: 'FeatureCollection',
+    MULTI_PREFIX: 'Multi',
+    MULTI_POINT: 'MultiPoint',
+    MULTI_LINE_STRING: 'MultiLineString',
+    MULTI_POLYGON: 'MultiPolygon'
   },
   modes: {
     DRAW_LINE: 'draw_line_string',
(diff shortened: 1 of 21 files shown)


src/constants.js
```diff
@@ -1,12 +1,21 @@
...
-  CONTROL_BUTTON_LINE_CLASS: 'mapbox-gl-draw_line',
-  CONTROL_BUTTON_POLYGON_CLASS: 'mapbox-gl-draw_polygon',
-  CONTROL_BUTTON_POINT_CLASS: 'mapbox-gl-draw_point',
-  CONTROL_BUTTON_TRASH_CLASS: 'mapbox-gl-draw_trash',
-  CONTROL_GROUP_CLASS: 'mapboxgl-ctrl-group',
-  ATTRIBUTION_CLASS: 'mapboxgl-ctrl-attrib',
-  ACTIVE_BUTTON_CLASS: 'active',
+  classes: {
+    CONTROL_BASE: 'mapboxgl-ctrl',
+    CONTROL_PREFIX: 'mapboxgl-ctrl-',
```
mcwhittemore: These two have the same value. We likely only need one.

mcwhittemore:
Hmm.... lots of constants. Generally I don't use constants except for event name. This is because events are points where my code interactions others code and it is very important for me to use a particular value.

Now, is using them all the time bad... I don't really know. I guess the question is what one is trying to achieve with them. What do these constants add to Draw? For instance, what positives does defining the GeoJSON types create? Does it fix bugs? Does it simplify assigning values? Does it make reviewing the code easier? And what are the negatives?

Personally I think it makes it harder to define values as there is a lot more to type. I'm on the fence about if its good for reviewing. On one hand there is a clear list of values in the code, on the other JS doesn't ensure the attribute is in that file so it might just be undefined...

I do think that the constants for feature properties is a good idea as its a clear list we could check for when processing the features inside of the mode_handler.

What are your thoughts? What are you hoping to get out of a lot of constants?

src/constants.js
```diff
@@ -1,12 +1,21 @@
...
+    CONTROL_BUTTON_POINT: 'mapbox-gl-draw_point',
+    CONTROL_BUTTON_TRASH: 'mapbox-gl-draw_trash',
+    CONTROL_GROUP: 'mapboxgl-ctrl-group',
+    ATTRIBUTION: 'mapboxgl-ctrl-attrib',
+    ACTIVE_BUTTON: 'active',
+    BOX_SELECT: 'mapbox-gl-draw_boxselect'
+  },
+  sources: {
+    HOT: 'mapbox-gl-draw-hot',
+    COLD: 'mapbox-gl-draw-cold'
```
mcwhittemore: how many places are these used?

src/constants.js
```diff
@@ -19,6 +28,17 @@ module.exports = {
     LINE: 'line_string',
     POINT: 'point'
   },
+  geojsonTypes: {
```
mcwhittemore: How many places are these used?

src/constants.js
```diff
@@ -38,5 +58,14 @@ module.exports = {
...
+  },
+  meta: {
+    FEATURE: 'feature',
+    MIDPOINT: 'midpoint',
+    VERTEX: 'vertex'
+  },
+  activeStates: {
+    ACTIVE: 'true',
+    INACTIVE: 'false'
   }
```
mcwhittemore: It would be nice to know that `meta` and `activeStates` were related to render features.

davidtheclark:
> What are your thoughts? What are you hoping to get out of a lot of constants?

This is one of those situations in which I'd be happy to accept The Voice of Authority & Experience, if anyone wants to claim that voice. (シ_ _)シ

But if I were to try to articulate some opinions myself, here are my current thoughts:

Constants are preferable to magic numbers pretty much all the time. Unlike magic numbers, string literals usually kind of name themselves; so that benefit isn't so important. But these might be benefits to constants even with string literals:
- Provide a name when string literals actually don't really name themselves and are kind of "magic", like `"true"` and `"false"` for `active` states in Draw.
- If you're going to use a value multiple times, write it down and store it in one place.
- Look at the strings side-by-side and recognize inconsistencies you might otherwise miss (e.g. `mapboxgl-foo` vs `mapbox-gl-foo`).
- Avoid bugs when the strings need to correspond to something else that even run-time errors might not catch, like CSS selectors.
- When static analysis gets involved, you could get some of the same benefits that enums provide in typed languages (pre-runtime errors).

Downsides:
- A little extra typing. But I don't think that's much of an issue, really.
- Another convention you feel like you have to manually enforce or else let slide into inconsistent usage. Might make you feel pedantic, which sucks.
- Additional indirection that might not always seem worth it (having to look in a different file for the constant name). Sometimes this could be reduced by using local constants where they don't need to be shared (e.g. I think Draw's class names might only be used in `ui.js`).

Those are my current thoughts.

src/constants.js
tmcw: The second ends with a -, the first doesn't

tmcw:
I like @davidtheclark's reasoning here - this PR is definitely testing the idea of taking constants to the max, and I don't think it yet runs into any of the major downsides. The code increase is minimal - +49 lines.

And repeating/adding three points:

* Once you add Flow into the equation, the benefit of turning typos from runtime errors like "LineString" === "Linestring" to compile-time errors like `constants has no property Linestring` is pretty great.
* Constants provide a nice path for making CSS classes configurable in the future.
* This PR already provides the immediate benefit of making tests more robust; since we're testing for CSS class existence, before we would have to change the class name in two places, now we only have to change it in one.

I think that this PR is a useful experiment and I believe the early results & potential future benefits are solid. 👍 from me.

src/constants.js
mcwhittemore: oh, wow. Maybe a good difference between constants and string literals is that they can be easier to distinguish.

mcwhittemore:
Hmmm. OK. I'm up for seeing what lots of constants can do for us. Can we follow this PR up with some linting and maybe some flow work? I'd really like to know that the constant I'm using is defined by the constants file. I don't see a rule for this in eslint, but maybe flow would catch it?

Issue #289, opened 2016-04-22
Custom Modes


mcwhittemore:
The modularity provided by a mode system should be exposed to end-users so they can replace out the default interactions with their own.

drewbo:
@mcwhittemore any ideas on how to implement this? It seems pretty tricky because the modes are so closely tied to some of the api methods ala https://github.com/mapbox/mapbox-gl-draw/blob/master/src/api.js#L129 instead of falling back to the larger mode modules. Happy to kick around some ideas + code as we're looking to use this again in another project with some custom behavior

drewbo:
Actually, the more I look; the file I mentioned is really the only place where the modularity of the mode system breaks down a little (nice job!). Would the next steps to implementing this feature be something like:
- Abstract the API checks for other custom modes
- Document required methods for a custom mode module/object
- Provide a method or option for loading those custom modes

mcwhittemore:
@drewbo sorry for just seeing this.

Those bullet points seem right to me.

We'll want to make sure the API for these modes is really pretty solid. Asking people to write custom modes and then breaking the API a few times won't be good. My key concern is cleaning up how events are handled/registered.

mcwhittemore:
Here is a ticket where @davidtheclark and I are talking about this problem - https://github.com/mapbox/mapbox-gl-draw/issues/459

BhavanPatel:
@mcwhittemore @drewbo I think GeoJSON support rectangle... as I have seen in this [Demo](http://geojson.io/) and map seems mapbox.

mcwhittemore:
@BhavanPatel GeoJSON supports Polygons and thus you can create GeoJSON rectangles. My comments in #604 aren't trying to say this isn't possible, but that Rectangles are out of the scope of Draw. It should be noted, that Draw used to have Rectangle support and it caused unexpected functionality multiple times so it was removed. Around this same time, we decided that Draw would only support the code [GeoJSON Spec](https://tools.ietf.org/html/rfc7946) and that custom modes could provide for extra functionality such as Rectangles and Circles.

Personally, I really want a mode that lets me Draw hexagons. Sadly I don't have a lot of time for this project currently. @drewbo has shown interest in getting Custom Modes to happen, but more help from others in the community would be great too.

PR #452, opened 2016-07-05
Option to be able to drag the feature in direct_select mode

kiRach:
Hi!

We use this library a lot for our project: https://compstak.com

And recently we released a new version with Google maps replaced by Mapbox GL maps. Everything works like a charm, thanks for your work. But there is one issue and we have a couple users that already complained about it - so it's pretty important for us.

Little story: we use maps for search. And users are able to use different map search tools. One of these tools is so called "polygon search": user just puts a polygon shape to the map and should be able to resize/drag it to perform search. Like this: 

like this

But the problem is that your library has separate modes for dragging features (simple_select) and for resizing (direct_select). So it's not possible for us to allow users to drag and to resize polygon at the same time. (And it used to work when we had goole maps - so that's why users complain about it).

Here is my attempt to solve this issue by creating new mode. I called if `mixed` mode (but you can suggest better name). Basically it's just a mix of `simple_select` and `direct_select` modes. 

The work is still in progress, and I have to write some tests at least. But just let me know what do you think about idea of this mode. And what else I should implement here.

Thanks!


mcwhittemore:
@kiRach This is great. Thanks. Also, nice use of custom styles. 😄 

It looks like the main goal here is to add dragging a feature to `direct_select` mode. Is that correct?

`mixed` is a bit confusing of a name. Maybe `single_feature_select`?

I think the next step here is to refactor the shared functionality with `direct_select` and `simple_select` out into shared files.

kiRach:
@mcwhittemore I've just finished work on this feature. Review please.

I decided that it doesn't worth to create a separate mode for dragging feature because it's basically the `direct_mode` with little changes. So instead of separate mode I just added the option `dragFeature` to direct select mode. Also added some tests for the cases when `dragFeature` is enabled and disabled for `direct_select` mode.

Let me know what you think. If it's ok  I'll update the README to add the option description.

mcwhittemore:
@kiRach I'll give this a good look over tomorrow

src/modes/direct_select.js
```diff
@@ -73,33 +79,67 @@ module.exports = function(ctx, opts) {
...
+
+      if (dragFeature) {
+
+        this.on('mousedown', CommonSelectors.isActiveFeature, function (e) {
+          dragMode = 'feature';
+
+          ctx.map.dragPan.disable();
+          this.render(e.featureTarget.properties.id);
+
+          // Set up the state for drag moving
```
mcwhittemore: We should call `startDragging()` here.

src/modes/direct_select.js
```diff
@@ -73,33 +79,67 @@ module.exports = function(ctx, opts) {
...
-
-        var selectedCoords = selectedCoordPaths.map(coord_path => feature.getCoordinate(coord_path));
-        var selectedCoordPoints = selectedCoords.map(coords => ({
-          type: Constants.geojsonTypes.FEATURE,
-          properties: {},
-          geometry: {
-            type: Constants.geojsonTypes.POINT,
-            coordinates: coords
+
+        if (dragFeature && dragMode === 'feature') {
```
mcwhittemore: Shouldn't the `dragMode` tests be enough?

mcwhittemore:
@kiRach I liked this a bit more when it was its own mode and not a flag in a mode.

@davidtheclark and I have been talking about dropping the options argument in `changeMode` to help remove state for the modes so this PR is stepping into that convo a bit. https://github.com/mapbox/mapbox-gl-draw/issues/397#issuecomment-228125969

It is also stepping into a convo we've been having about changing how mode handlers work.
https://github.com/mapbox/mapbox-gl-draw/issues/459

I'm pretty comfortable with adding extra modes as I want to [create a way for users to create custom modes](https://github.com/mapbox/mapbox-gl-draw/issues/289) and having non-default modes as part of the tool before that happens mean we would have a few custom modes automatically when we roll out that feature.

That said... this is really close to `direct_select` and I'd be lying if I said I'm not tempted to make this the default behavior.

kiRach:
@mcwhittemore The main issue with separate mode was that it was almost 90% of copy-paste from `direct_select` mode. And it's hard to refactor and extract common functionality a lot of functions depend on mode state (variables which are declared in the in mode function) and context dependent (for example mode_handler call the start method of mode passing another context for `this`). So I decided it's better to implement it using option for `direct_select` mode. 

If you want - I can move back to separate mode, but it still will be a lot of copy-paste from `direct_select` mode. Or I can change the PR to make dragging the feature default for `direct_mode`. Let me know. And thanks for review!

ilyamilosevic:
Interesting! Recently I had almost the same task. To avoid problem with modes I implemented a plugin similar to [Leaflet AreaSelect](https://github.com/heyman/leaflet-areaselect), but with ability to change hole.
map_filter

davecranwell:
+1 for seeing this PR completed. I too use Draw to aid geo-bounded searches, and the draggability issue in the GL version, compared to the old (e.g on geojson.io) is definitely something which would greatly improve usability.

mcwhittemore:
This has been implemented as the default behaviour via #536. It will be out in the 0.15.0 release.

PR #453, opened 2016-07-05
add marker to make click to end polygon clear

### LINKED ISSUE #448: make it easier to finish polygons and lines
Right now double click on first/last point or return finish polygons. Fundamentally this is solid. There is a problem. The hit area for first/last point are tiny. I see two improvements:

While drawing, create circle features at each node and use them to give first/last point bigger hit area.

While hovering over first/last feature, change the cursor to something that communicates: "finish drawing".

mcwhittemore:
Closes #448

I still need to fix and write some tests, but @samanpwbb what do you think of this functionality.

easy end

### DIFF
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -28,21 +28,18 @@ var ModeHandler = function(mode, DrawContext) {
   var delegate = function (eventName, event) {
     var handles = handlers[eventName];
     var iHandle = handles.length;
-    var handlerCalled = false;
     while (iHandle--) {
       var handle = handles[iHandle];
       if (handle.selector(event)) {
         handle.fn.call(ctx, event);
-        handlerCalled = true;
+        DrawContext.store.render();
+        DrawContext.ui.updateMapClasses();
+
         // ensure an event is only handled once
         // we do this to let modes have multiple overlapping selectors
         // and relay on order of oppertations to filter
         break;
       }
-    }
-    if (handlerCalled) {
-      DrawContext.store.render();
-      DrawContext.ui.updateMapClasses();
     }
   };
 
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -34,6 +34,10 @@ var ModeHandler = function(mode, DrawContext) {
       if (handle.selector(event)) {
         handle.fn.call(ctx, event);
         handlerCalled = true;
+        // ensure an event is only handled once
+        // we do this to let modes have multiple overlapping selectors
+        // and relay on order of oppertations to filter
+        break;
       }
     }
     if (handlerCalled) {
--- a/src/modes/draw_polygon.js
+++ b/src/modes/draw_polygon.js
@@ -38,18 +38,16 @@ module.exports = function(ctx) {
       doubleClickZoom.disable(ctx);
       ctx.ui.queueMapClasses({ mouse: Constants.cursors.ADD });
       ctx.ui.setActiveButton(Constants.types.POLYGON);
-      this.on('mousemove', CommonSelectors.isVertex, () => {
-        ctx.ui.queueMapClasses({ mouse: Constants.cursors.POINTER });
-      });
-      this.on('click', CommonSelectors.isVertex, () => {
-        polygon.removeCoordinate(`0.${currentVertexPosition}`);
-        currentVertexPosition--;
-        return ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [polygon.id] });
-      });
       this.on('mousemove', CommonSelectors.true, e => {
         polygon.updateCoordinate(`0.${currentVertexPosition}`, e.lngLat.lng, e.lngLat.lat);
+        if (CommonSelectors.isVertex(e)) {
+          ctx.ui.queueMapClasses({ mouse: Constants.cursors.POINTER });
+        }
       });
       this.on('click', CommonSelectors.true, handleClick);
+      this.on('click', CommonSelectors.isVertex, () => {
+        return ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [polygon.id] });
+      });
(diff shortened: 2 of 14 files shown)


mcwhittemore:
meh - blocked by this - https://github.com/Turfjs/turf/issues/429

mcwhittemore:
@davidtheclark for the review

davidtheclark:
@mcwhittemore We should probably apply the same treatment to lines, as well, right? Or did you want to save that for another PR?

mcwhittemore:
@davidtheclark yea, I'm not sure if we should add it to lines. Lines don't end. I guess we could put a point at the last placed vertex...

src/modes/draw_polygon.js
```diff
@@ -41,7 +37,17 @@ module.exports = function(ctx) {
       doubleClickZoom.disable(ctx);
       ctx.ui.queueMapClasses({ mouse: Constants.cursors.ADD });
       ctx.ui.setActiveButton(Constants.types.POLYGON);
-      this.on('mousemove', CommonSelectors.true, handleMouseMove);
+      this.on('mousemove', CommonSelectors.isVertex, () => {
+        ctx.ui.queueMapClasses({ mouse: Constants.cursors.POINTER });
+      });
+      this.on('click', CommonSelectors.isVertex, () => {
+        polygon.removeCoordinate(`0.${currentVertexPosition}`);
+        currentVertexPosition--;
```
davidtheclark: I've been curious about this line: why do we have it if we exit the mode immediately after?

davidtheclark:
> I guess we could put a point at the last placed vertex...

That's what I was thinking, since that's how you end the line — by clicking the last vertex again.

src/modes/draw_polygon.js
```diff
@@ -89,9 +96,28 @@ module.exports = function(ctx) {
...
+          properties: {
+            meta: Constants.meta.VERTEX,
+            parent: polygon.id,
+            coord_path: '0.0'
+          },
+          geometry: {
+            type: Constants.geojsonTypes.POINT,
+            coordinates: geojson.geometry.coordinates[0][0]
+          }
+        });
```
davidtheclark: Should this use the `create_vertex.js` function?

test/draw_line_string.test.js
```diff
@@ -137,7 +137,7 @@ test('draw_line_string render active line with 1 coordinate', t => {
     },
     geometry: {
       type: 'LineString',
-      coordinates: [[0, 0]]
+      coordinates: [[0, 0], [15, 15]]
```
davidtheclark: Doesn't this make the test description inaccurate, since we have 2 coordinates?

test/draw_line_string.test.js
```diff
@@ -151,7 +151,7 @@ test('draw_line_string render active line with 1 coordinate', t => {
...
```
davidtheclark: Same here: does test description need changing?

davidtheclark:
I really like the interaction: makes it very easy to close the polygon.

test/draw_line_string.test.js
mcwhittemore: Oh, this test is just wrong? It should return nothing right, as its an invalid LineString?

davidtheclark: If it was failing I think that's because of your change at https://github.com/mapbox/mapbox-gl-draw/pull/453/files#diff-7a0124268c46d39459f7595fd9d4d1caR83. The test was correct for the prior behavior, but you're changing that, right?

mcwhittemore: Yea, it was rendering an invalid LineString before. This happens in tests more than it does in real life as few people can click without moving the mouse and mousemove adds the second position.

mcwhittemore:
@davidtheclark I just realized I forgot to write tests for this new interaction (yay for coverage). Sadly it means converting the `draw_polygon` tests to be able to work with render... Part of me wants to drop usage of `queryRenderedFeatures` for finding what is being clicked on.

davidtheclark:
> converting the draw_polygon tests

Do you need to convert old tests or just write new ones?

mcwhittemore:
> Do you need to convert old tests or just write new ones?

I was able to just add a test that was render dependent. I got a bit confused about which context the tests were using.

src/modes/draw_polygon.js
mcwhittemore: Yea... I'll look into this more. My first pass didn't have it but than there was an extra point.

test/draw_polygon.test.js
```diff
@@ -563,6 +580,31 @@ test('draw_polygon interaction', t => {
       st.end();
     });
 
+    // FIVE CLICK TEST
+
+    // This test would require this render to be operational.
```
davidtheclark: I'm confused by this comment. Is the test invalid in its current form

mcwhittemore: Nope. I forgot to remove this when I fixed the test.

davidtheclark:
This PR seems good to me now, right? You have a test for the new vertex appearing, and a test for closing the vertex. Merge? Or did you have anything else in mind?

mcwhittemore:
Yea, I'm going to merge this and than make an issue about ending LineStrings. I don't know how I feel about it.

mcwhittemore:
@davidtheclark just realized the issue is asking for what you are asking for so I'm going to add this to line strings and add the second point.

mcwhittemore:
@davidtheclark - ok, I think this is ready for you for the last time.

src/lib/mode_handler.js
```diff
@@ -34,6 +34,10 @@ var ModeHandler = function(mode, DrawContext) {
       if (handle.selector(event)) {
         handle.fn.call(ctx, event);
         handlerCalled = true;
+        // ensure an event is only handled once
+        // we do this to let modes have multiple overlapping selectors
+        // and relay on order of oppertations to filter
+        break;
```
davidtheclark: If we are limiting to one handler I don't think we need the handlerCalled tracking, right?

PR #454, opened 2016-07-06
Expose Draw.modes; closes #267

### LINKED ISSUE #267: Draw modes should be an ENUM
> Would prefer modes to be either enums (Draw.modes.DEFAULT) or separate methods (Draw.startDirectSelect(id), Draw.startDrawing(Draw.types.LINE)) although then I'm not positive what draw.changemode should emit.

> - @drewbo

davidtheclark:
Ref #267.

In the process of doing this I thought it would be a good idea to change `DRAW_LINE` to `DRAW_LINE_STRING`, for consistency with geojson types.

@mcwhittemore for review.


mcwhittemore:
This looks good to me.

PR #455, opened 2016-07-06
Upgrade dependencies, lint tests

davidtheclark:
I upgraded dependencies, made ESLint check test, and fixed the few linting errors it found.

I replaced `turf-area` with `geojson-area` because it looks like `geojson-area` is all we need. Please correct me if I'm wrong!

@mcwhittemore for review.


mcwhittemore:
Rebased. Once the tests pass, lets merge.

Issue #372, opened 2016-06-16
You should not be able to draw a feature past the north and south polls


mcwhittemore:
![image](https://cloud.githubusercontent.com/assets/302992/16016607/b814a628-3150-11e6-8016-150d98c6f2a5.gif)

mcwhittemore:
There is also a problem with dragging a feature around the world.

![image](https://cloud.githubusercontent.com/assets/1551510/16627802/d7247fbe-437c-11e6-86ce-dcc26baa02a5.png)

PR #457, opened 2016-07-07
Move feature limits

davidtheclark:
Ref #372.

This just addresses north-south movement towards the poles (not east-west movement more than 360), because that was enough for one PR.

@mcwhittemore I think this satisfies the requirements we discussed earlier today. From playing around with various scenarios, I believe we're now unable to 

- move a point past the poles such that it starts warping, or
- move any feature outside the projection such that I'm unable to re-select it

Ready for review.


mcwhittemore:
@davidtheclark this is looking good. I'll play with it a bunch in the morning and make sure there aren't any edge cases you're missing.

davidtheclark:
I am trying to work in longitude adjustments before the end of the night ... we'll see ...

davidtheclark:
This was a good workout for my mathless brain.

I think I have it working now so that you are stopped north-south at the poles and can drag endlessly east-west.

I did notice one bug that I don't think needs to be entangled with this particular functionality: https://github.com/mapbox/mapbox-gl-draw/issues/458

davidtheclark:
A couple of notes for reviewing:

- The changes to `interaction_events.test.js` (which were a huge pain to figure out) were necessary because when I first wrote those tests I wasn't thinking about the +/-90 limits of latitude (noob) and created features that exceed them. That was fine before, but with the changes in the PR that isn't fine anymore (and I shouldn't have done it in the first place).
- I ran into a situation where if you've dragged a feature such that a vertex is off the projection, trying to create a midpoint related to it will error in gl-js. So I thought it'd fine to just not create midpoints if one vertex is off the projection.

mcwhittemore:
This looks great! Merged.

Issue #459, opened 2016-07-07
Audit one-handler-per-event system


davidtheclark:
I'm a little concerned that there might be some reason to take another look at the move in https://github.com/mapbox/mapbox-gl-draw/pull/453/commits/1d4527b5bf33330771d2a4be167722c52c00359c to enforce only one handler per event.

Maybe there's nothing wrong, but here are my concerns:
1. I'm not sure we're currently using the events in a way that expects this everywhere. For example, I wonder if https://github.com/mapbox/mapbox-gl-draw/blob/master/src/modes/direct_select.js#L78 is now subtly wrong, because if its handler fires it means the other `mousedown` handlers won't, right? — but they should.
2. Generally I just kind of find this more confusing than the other way, because we are accustomed to `addEventListener` and EventEmitters where you can attach as many handlers as you want, independently. Why rely on order of operations to filter _all_ handlers when you could rely on explicit conditionals to filter those handlers you want to filter?

So because of the first point, I opened this issue because at some point the code should be reviewed to ensure there isn't weirdness around expecting more handlers to be called than will be. (The second point is just my question.)

mcwhittemore:
@davidtheclark I agree its a bit weird at first I tried to replace the `CommonSelectors.true` with `CommonSelectors.noTarget` but that failed. Looking at it now, we might be able to replace `CommonSelectors.true` with `CommonSelectors.notVertex` or something like that.

A bit of history here, events have only been able to be handled once per mode since 390585ae307b8cb29e13aaa1c762534cf7566624. I missed you're removal of the break in ae730111a9965342401d873369106f3871eb188e when I reviewed it, so I was concerned there were more places where two events were firing. Did you do check that more handlers weren't being fired when you removed the break?

I think the best approach would be to drop the selector filter from mode_handler and have normal event handlers per mode. The common selectors can still be used, but they will be used in if statements in the event handler not before the event handler.

davidtheclark:
> I think the best approach would be to drop the selector filter from mode_handler and have normal event handlers per mode. The common selectors can still be used, but they will be used in if statements in the event handler not before the event handler.

That sounds like a good approach to me. The closer we get to the common event handler behavior, the better.

mcwhittemore:
There is an extra handler in `direct_select` that needs to be deleted.

https://github.com/mapbox/mapbox-gl-draw/blob/696a2078143ac2ea314698b1351c354236be7365/src/modes/direct_select.js#L141

[...]

PR #463, opened 2016-07-11
[WIP] Simplify mode interface

mcwhittemore:
@davidtheclark I took a quick stab at dropping selectors per #459, but I don't really like what it does to the code.

I might be ok with this if the exported object just exposed `onClick` rather than setting up click handlers in the exposed `start` function.

What do you think?

### DIFF
--- a/src/lib/common_selectors.js
+++ b/src/lib/common_selectors.js
@@ -40,6 +40,12 @@ module.exports = {
     if (!featureTarget.properties) return false;
     return featureTarget.properties.meta === Constants.meta.VERTEX;
   },
+  isMidpoint: function(e) {
+    var featureTarget = e.featureTarget;
+    if (!featureTarget) return false;
+    if (!featureTarget.properties) return false;
+    return featureTarget.properties.meta === Constants.meta.MIDPOINT;
+  },
   isShiftDown: function(e) {
     if (!e.originalEvent) return false;
     return e.originalEvent.shiftKey === true;
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -11,14 +11,11 @@ var ModeHandler = function(mode, DrawContext) {
   };
 
   var ctx = {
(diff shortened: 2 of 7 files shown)


davidtheclark:
> I don't really like what it does to the code.

Could you describe what you don't like? Is it that the handler functions are longer and all broken up with conditionals (I kind of don't like that, myself)? If so, one way out of that could be to move the real logic for each possible response into its own function and then make the handler nothing more than a set of conditionals determining which response function(s) to call. 

Does that make sense? So this:

```js
this.on('keyup', e => {
  if(CommonSelectors.isEscapeKey(e)) {
    ctx.store.delete([line.id], { silent: true });
    return ctx.events.changeMode(Constants.modes.SIMPLE_SELECT);
  }

  if (CommonSelectors.isEnterKey(e)) {
    return ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [line.id] });
  }
});
```

Could become this:

```js
function handleEscape() {
   ctx.store.delete([line.id], { silent: true });
   ctx.events.changeMode(Constants.modes.SIMPLE_SELECT);
}

function handleEnter() {
  ctx.events.changeMode(Constants.modes.SIMPLE_SELECT, { featureIds: [line.id] });
}

this.on('keyup', e => {
  if (CommonSelectors.isEscapeKey(e)) return handleEscape();
  if (CommonSelectors.isEnterKey(e)) return handleEnter();
});
```

Does that seem like an improvement?

> I might be ok with this if the exported object just exposed onClick rather than setting up click handlers in the exposed start function.

I think I really like this idea. It would mean we'd be passing functions around instead of using magic strings. Sounds great.

mcwhittemore:
@davidtheclark yea, your suggestion is the best but it moves us further away from having a mode be an object rather than a function. I need to think more.

mcwhittemore:
Going to go with https://github.com/mapbox/mapbox-gl-draw/pull/466

PR #466, opened 2016-07-15
Modes as Classes

mcwhittemore:
@kiRach and @davidtheclark 

After @davidtheclark's and my conversations about cleaning up modes and @kiRach observations of how verbose new modes can be I think we should move to make modes into classes. This will let them be extended (so you can make slightly different modes without much work) and will standardize how the mode interfaces with the rest of Draw a bit better.

This PR is just a POC. It does not work at all.

### DIFF
--- a/src/feature_types/feature.js
+++ b/src/feature_types/feature.js
@@ -1,16 +1,11 @@
 var hat = require('hat');
 var Constants = require('../constants');
 
-var Feature = function(ctx, geojson) {
-  this.ctx = ctx;
+var Feature = function(geojson) {
   this.properties = geojson.properties || {};
   this.coordinates = geojson.geometry.coordinates;
   this.id = geojson.id || hat();
   this.type = geojson.geometry.type;
-};
-
-Feature.prototype.changed = function() {
-  this.ctx.store.featureChanged(this.id);
 };
 
 Feature.prototype.incomingCoords = function(coords) {
--- a/src/feature_types/feature.js
+++ b/src/feature_types/feature.js
@@ -50,4 +50,4 @@ Feature.prototype.internal = function(mode) {
   };
 };
 
-export default Feature;
+module.exports = Feature;
--- a/src/feature_types/feature.js
+++ b/src/feature_types/feature.js
@@ -55,4 +55,4 @@ Feature.prototype.internal = function(mode) {
   };
 };
 
-module.exports = Feature;
+export default Feature;
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -7,7 +7,7 @@ const createVertex = require('../lib/create_vertex');
 import ModeInterface from './mode_interface';
 
 export default class DrawLineStringMode extends ModeInterface {
-  constructor(store, ui) {
+  constructor(options, store, ui) {
     super();
     this.line = new LineString({
       type: Constants.geojsonTypes.FEATURE,
@@ -64,7 +64,7 @@ export default class DrawLineStringMode extends ModeInterface {
     this.changeMode(Constants.modes.SIMPLE_SELECT);
   }
 
-  changeMode(nextModeName, store, ui) {
+  onChangeMode(nextModeName, store, ui) {
     this.doubleClickZoom(true);
     ui.setActiveButton();
     if (store.get(this.line.id) === undefined) return;
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -6,7 +6,7 @@ const Constants = require('../constants');
 const createVertex = require('../lib/create_vertex');
 
 module.exports = function(ctx) {
-  const line = new LineString(ctx, {
+  const line = new LineString({
     type: Constants.geojsonTypes.FEATURE,
     properties: {},
     geometry: {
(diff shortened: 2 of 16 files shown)


kiRach:
@mcwhittemore Cool! Exactly what I wanted to propose. Modes should be classes definitely. Also may be it would be nice to have some basic mode which could be extended by inheritance/HOC to implement custom mode.

mcwhittemore:
> it would be nice to have some basic mode which could be extended

@kiRach this is what I'm thinking the `ModeInterface` would be. Are you thinking it should have more functionality than this?

src/feature_types/feature.js
```diff
@@ -55,4 +50,4 @@ Feature.prototype.internal = function(mode) {
   };
 };
 
-module.exports = Feature;
+export default Feature;
```
davidtheclark: Curious: why switch to using ES2015 exports but still CommonJS `require` instead of `import`?

mcwhittemore: My plan is to be moved over to import by the end of the this pr.

src/modes/draw_line_string.js
```diff
@@ -1,96 +1,101 @@
...
-  const line = new LineString(ctx, {
-    type: Constants.geojsonTypes.FEATURE,
-    properties: {},
-    geometry: {
-      type: Constants.geojsonTypes.LINE_STRING,
-      coordinates: []
+import ModeInterface from './mode_interface';
+
+export default class DrawLineStringMode extends ModeInterface {
+  constructor(store, ui) {
```
davidtheclark: I might favor passing an object as the argument to this and other methods, instead of an argument list. That way the argument lists won't get long and it will be easier to add or change down the line as needed. What do you think?

src/feature_types/feature.js
mcwhittemore: Soon I want to classify the features too. Going with classes makes sense and if we're going to go with classes, we might as well do import vs require.

src/modes/draw_line_string.js
```diff
@@ -1,96 +1,101 @@
...
+      }
+    });
+
+    this.currentVertexPosition = 0;
+
+    store.add(this.line);
+    store.clearSelected();
+    this.doubleClickZoom(false);
+    ui.queueMapClasses({ mouse: Constants.cursors.ADD });
+    ui.setActiveButton(Constants.types.LINE);
```
davidtheclark: So it may be inevitable if we want to allow for custom modes, but this kind of unfortunately balloons the scope of the public API, since now everything in `store` and `ui`, and anything else that gets passed into these methods, is public API, right?

davidtheclark: *Or* would we say that custom modes don't get these arguments that expose internals?

src/modes/draw_line_string.js
```diff
@@ -1,96 +1,101 @@
...
+  }
+
+  onMousemove(e) {
+    this.line.updateCoordinate(this.currentVertexPosition, e.lngLat.lng, e.lngLat.lat);
```
davidtheclark: Should this only happen after you've clicked?

src/modes/draw_line_string.js
mcwhittemore: yea... we will want to think about how exposing this stuff works with custom modes. For now, the custom modes are being shipped with Draw so I'm fine with this.

src/modes/draw_line_string.js
mcwhittemore: If anything, I want fewer arguments as we work on Draw not more. I think we can handle your concerns about exposing the internals by adding more default functions to the mode like `this.setSelected` to let the mode set the selected features.

davidtheclark: Using an object, then, would allow you to delete an argument without changing the signature (e.g. if we decided to no longer pass the first argument into a method).

> I think we can handle your concerns about exposing the internals by adding more default functions to the mode like this.setSelected to let the mode set the selected features.

I think that is a good idea. Are you thinking we'd then *not* pass the arguments to custom modes, and expect them to rely on the class's methods?

davidtheclark:
@mcwhittemore Looks like a good start to me. I was thinking about what having classes-matching-an-interface offered as a benefit over the current pattern of functions that return objects-matching-an-interface --- and I think that question has a good answer in your suggestion of exposing methods on the base class which custom modes can then use to their ends (instead of needing access to Draw internals).

src/feature_types/feature.js
davidtheclark: This makes sense to me. It's for browser usage so will be compiled anyway, so we may as well use all ES2015. 

I like that you were able to drop the `ctx` from these features.

src/modes/draw_line_string.js
mcwhittemore: > Are you thinking we'd then not pass the arguments to custom modes, and expect them to rely on the class's methods?

Yea, except that I don't want to make a distinction between default modes and custom ones.

src/modes/draw_line_string.js
mcwhittemore: - [ ] to do

src/modes/draw_line_string.js
davidtheclark: So, to clarify, the method arguments would be removed altogether, right?

mcwhittemore: That is what I'm envisioning, but it's also pretty limiting and I don't want to replicate the full api of store and ui in the mode so I'm not sure yet.

mcwhittemore:
I think this PR is DOA. I really like this idea, but it is going to take time that I don't have right now. Closing for now.

PR #467, opened 2016-07-21
Prevent deleted features emitting twice

davidtheclark:
We ran into a bug in which the `draw.delete` event was firing twice before the next line of code, which clears the array of features to emit, was cleared. So we ended up with listeners thinking the same feature was deleted twice.

My guess as to why this was happening is because the firing of the event sent us down a completely synchronous code path that ended up looping us back to that very spot where the event fired. Then the event fired again before we got to that next line that clears the array. Solution, then, is to clear the array before firing the event.

cc @mcwhittemore

### DIFF
--- a/src/lib/throttle.js
+++ b/src/lib/throttle.js
@@ -16,10 +16,10 @@ function throttle(fn, time, context) {
       args = arguments;
 
     } else {
-      // call and lock until later
+      // lock until later then call
+      lock = true;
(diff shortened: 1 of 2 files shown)


davidtheclark:
Added another change to lock `render` within `throttle` *before* invoking it, so it cannot call itself synchronously and thereby avoid the throttling.

PR #469, opened 2016-07-22
Use function expression instead of function declaration

SamSamskies:
This fixes a scoping issue with these functions when using Webpack. These two functions are undefined at runtime when they are defined using function declarations. It works fine when defined using function expressions. I can't figure out why this is happening, but it won't hurt to merge this change.


tmcw:
This sounds like a webpack bug: why would these functions be undefined as function declarations which are hoisted?

I'm 👎 on merging this: if there's a bug, it's in webpack and should be fixed there, rather than worked-around here.

PR #470, opened 2016-07-25
Fire draw.update on mouseout when dragging

davidtheclark:
This PR introduces `mouseout` events and causes a `draw.update` event to fire if you've been dragging a feature or vertex and your mouse leaves the container.

Before this, we had a problem that you could drag-move something until your mouse leaves the container, release the mouse button outside, then return, and the feature would have changed its position but no` draw.update` event would have been fired. With these changes, a `draw.update` will be fired anytime you leave. (If you continue to hold the mouse button down and move back in, continuing the drag, then mouseup, ending the drag, you'll get another `draw.update` event.)

@mcwhittemore for review.

### DIFF
--- a/API.md
+++ b/API.md
@@ -314,7 +314,7 @@ Fired when one or more features are updated. The following will trigger this eve
 - `action: 'move'`
   - Finish moving one or more selected features in `simple_select` mode. The event will only fire when the movement is finished — i.e. when the user releases the mouse button or hits Enter.
 - `action: 'change_coordinates'`
-  - Finish moving one or more vertices of a selected feature in `direct_select` mode. The event will only fire when the movement is finished — i.e. when the user releases the mouse button or hits Enter.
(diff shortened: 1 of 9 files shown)


mcwhittemore:
@davidtheclark can we solve this a level up from the modes?

When mouseout happens we capture the state of the mouse. When mousein happens we compare the state from mouseout and fire a mousedown or mouseup if needed.

davidtheclark:
> When mousein happens we compare the state from mouseout and fire a mousedown or mouseup if needed.

Could try that. But the downside would be that you would have to return your mouse to the map before you get your `draw.update` event. For whatever interval of time between your mouse leaving the map and it re-entering (if you do), Draw's rendered map will reflect an update that has not triggered `draw.update`.

mcwhittemore:
hmmm... that is a good point. I'm good with emitting two updates for now.

davidtheclark:
Should I merge this? Or you want time to take a closer look?

PR #475, opened 2016-08-09
3 worlds

mcwhittemore:
@davidtheclark can you take this for a spin and tell me what you think?


davidtheclark:
aug-10-2016 11-21-21

I don't know @mcwhittemore ... this way of stopping the pan is pretty bumpy. As a user I would definitely think this was a bug, not deliberate.

Too bad we can't set the `maxBounds` option on a map via a Control.

mcwhittemore:
yea... that is what I'm thinking too.

Another option we could do is try to slowly pan back to the right zoom box, but that is going to take a lot more control.

mcwhittemore:
Closing this for now as it can be achieved via `maxBounds` when setting up the map object which happens outside of Draw.

Issue #472, opened 2016-08-08
Unable to style finished features


timiyay:
I'm using `mapbox-gl-draw` for a basic use case: 

1. users can draw a line to for measure between 2 points on the map
2. user clicks a `start` button to activate drawing mode
3. user draws their line(s)
4. during drawing, the lines are styled black
5. user double-clicks to finish their drawing
6. the feature switches styles, to green lines
7. the user can no longer edit the line

We are currently on v0.4.0 of this library, and are attempting to upgrade to the latest, but we're having trouble figuring out the new flow.

I'm able to get steps 1-5 working, but steps 6 and 7 are eluding me.

I'm unable to figure which mode I should be switching to, to achieve step 7. I'm trying to switch to `static` but, depending on my approach, it fails in a few ways:

1. if I'm listening for `draw.create` or `draw.modechange` events, and then call `changeMode('static')`, it causes an infinite loop. The event listener calls `changeMode('static')` which triggers the event listener which...

2. the line feature disappears (no styling).

What is the process for implementing this basic use case?

timiyay:
I'm able to achieve most of what I want, by copying the style examples from `test/fixtures/style_with_sources.json`.

In my case, I'm able to achieve my styling goals (steps 4 and 6 above) by adding the following styles:

```js
        // Lines: inactive (already drawn, not static)
        {
            "id": "gl-draw-line-active.cold",
            "type": "line",
            "filter": [
              "all",
              ["==", "$type", "LineString"],
              ["==", "active", "true"]
            ],
            "layout": {
              "line-cap": "round",
              "line-join": "round"
            },
            "paint": {
              "line-color": "#34C981",
              "line-width": 3
            },
            "interactive": true,
            "source": "mapbox-gl-draw-cold"
        },

        // Lines: active (being drawn)
        {
            "id": "gl-draw-line-active.hot",
            "type": "line",
            "filter": [
              "all",
              ["==", "$type", "LineString"],
              ["==", "active", "true"]
            ],
            "layout": {
              "line-cap": "round",
              "line-join": "round"
            },
            "paint": {
              "line-color": "#000",
              "line-dasharray": [0.2, 2],
              "line-width": 2
            },
            "interactive": true,
            "source": "mapbox-gl-draw-hot"
        }
```

If I remove `id`, nothing is rendered. If I remove `source`, then my custom styles are rendered _in addition to_ the default styles.

I'm finding this very confusing - the API docs clearly state:

[...]

PR #476, opened 2016-08-10
Update API docs regarding style ids and sources

timiyay:
This PR enhances the `API.md` docs to head off confusion for users who are learning to style `mapbox-gl-draw` features, as in this issue: https://github.com/mapbox/mapbox-gl-draw/issues/472

The must-have changes are the clarification of `source` and `id` here: https://github.com/mapbox/mapbox-gl-draw/compare/master...timiyay:feature/api-docs?expand=1#diff-88333ba11114e014e87198d04680144dR395

There's a few optional changes too:

* a second styling example - it's long example for styling for lines and polygons, which compliments the existing point-only example
* extra info on mode transitions, explaining how completed features always transition to `simple_select` (see https://github.com/mapbox/mapbox-gl-draw/issues/472#issuecomment-238578275)

I wasn't 100% on the vocabulary to use when explaining "completed" features. I was attempting to describe a feature that was in the process of being drawn by the user, who then double-clicked to "complete" the feature, and triggered the mode transition.

I also mimicked the use of **BOLD CAPITALISED** statements used in style source. These are arguably overkill, and I'd be happy to change them.

### DIFF
--- a/API.md
+++ b/API.md
@@ -41,6 +41,8 @@ Lets you select, delete and drag features.
 
 In this mode, features can have their active state changed by the user. To control what is active, react to changes as described in the events section below.
 
+Draw will transition into `simple_select` mode every time a single feature has completed drawing.
+
 ### `direct_select`
 
 `Draw.modes.DIRECT_SELECT === 'direct_select'`
@@ -382,13 +384,19 @@ Fired just after Draw calls `setData()` on `mapbox-gl-js`. This does not imply t
 
 Draw is styled by the [Mapbox GL Style Spec](https://www.mapbox.com/mapbox-gl-style-spec/) with a preset list of properties.
 
+`source`
+
 The `GL Style Spec` requires each layer to have a source. **DO NOT PROVIDE THIS** for styling draw.
 
-The `GL Style Spec` also requires an id. Draw will provide this for you. If you wish to set this id to interact with draw layers, know that Draw will add `hot` and `cold` if no source is provided.
-
-Draw moves features between sources for performance gains, because of this it is recommeneded that you **DO NOT** provide a source for a style despite the fact the `GL Style Spec` requires a source. **Draw will provide the source for you automatically**.
+Draw moves features between sources for performance gains, because of this it is recommended that you **DO NOT** provide a source for a style despite the fact the `GL Style Spec` requires a source. **Draw will provide the source for you automatically**.
 
 If you need to style gl-draw for debugging sources the source names are `mapbox-gl-draw-hot` and `mapbox-gl-draw-cold`.
+
+`id`
+
+The `GL Style Spec` also requires an id. **YOU MUST PROVIDE THIS**.
+
+Draw will add `.hot` and `.cold` to the end of your id.
 
 property | values | function
 --- | --- | ---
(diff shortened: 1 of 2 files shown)


API.md
```diff
@@ -382,14 +384,20 @@ Fired just after Draw calls `setData()` on `mapbox-gl-js`. This does not imply t
 
 Draw is styled by the [Mapbox GL Style Spec](https://www.mapbox.com/mapbox-gl-style-spec/) with a preset list of properties.
 
-The `GL Style Spec` requires each layer to have a source. **DO NOT PROVIDE THIS** for styling draw.
+`source`
```
mcwhittemore: I'd prefer **Source** over `source`.

API.md
```diff
@@ -382,14 +384,20 @@ Fired just after Draw calls `setData()` on `mapbox-gl-js`. This does not imply t
...
 
-The `GL Style Spec` also requires an id. Draw will provide this for you. If you wish to set this id to interact with draw layers, know that Draw will add `hot` and `cold` if no source is provided.
+The `GL Style Spec` requires each layer to have a source. **DO NOT PROVIDE THIS** for styling draw.
 
-Draw moves features between sources for performance gains, because of this it is recommeneded that you **DO NOT** provide a source for a style despite the fact the `GL Style Spec` requires a source. **Draw will provide the source for you automatically**.
+Draw moves features between sources for performance gains, because of this it is recommended that you **DO NOT** provide a source for a style despite the fact the `GL Style Spec` requires a source. **Draw will provide the source for you automatically**.
 
 If you need to style gl-draw for debugging sources the source names are `mapbox-gl-draw-hot` and `mapbox-gl-draw-cold`.
 
+`id`
```
mcwhittemore: I'd prefer **Id** over `id`.

API.md
```diff
@@ -398,11 +406,12 @@ mode |  simple_select, direct_select, draw_point, draw_line_string, draw_polygon
 
 Draw also provides a few more properties, but they should not be used for styling. For details on them, see `Using Draw with map.queryRenderFeatures`.
 
-### Example Custom Style
-
-With this style all Point features are blue and have a black halo when active. No other features are rendered, even if they are present.
+### Example Custom Styles
```
mcwhittemore: More examples are great. Please move the comments out of the code blocks and make sure you are always wrapping the style JSON in `mapbox.Draw({...})` so that reader understand its javascript vs json.

mcwhittemore:
@timiyay this is looking good. Do you think more examples in the docs are helpful? I'm worried they add a lot of clutter. Would a demo page be a better place for this?

timiyay:
@mcwhittemore I think more examples are helpful, and I'm open to moving them to an `EXAMPLES.md`.

Draw's architecture of modes, geometry types, and style filters is confusing, even for devs who're already familiar with the Mapbox Style Specification (our team). That isn't a criticism of Draw - it's an awesome plugin that saves us from rolling our own D3-based solution. It's dealing with a complex task, is all.

The current styling example doesn't address line or polygon features, which were my primary sources of our confusion. My extra example is adapted directly from our solution, so it provides people with a precedent they can use to guide their own line/polygon implementations.

I will update my PR accordingly, so we can see how this would look, and you can decide from there.

To repeat my previous comment - the clarification around the required `id` field is the critical part of this PR.

API.md
```diff
@@ -398,42 +406,9 @@ mode |  simple_select, direct_select, draw_point, draw_line_string, draw_polygon
...
-      'paint': {
-        'circle-radius': 5,
-        'circle-color': '#000088'
-      },
-      'interactive': true
-    }
-  ]
-});
-```
+See [EXAMPLES.md](https://github.com/mapbox/mapbox-gl-draw/blob/master/EXAMPLES.md) for examples of custom styles.reference.
```
timiyay: This links to `EXAMPLES.md` on `master`, so it'll be broken unless merged.

timiyay:
@mcwhittemore Would you like me to change this PR? Perhaps I could limit this PR to the doc updates you agree with, and move the extra stuff to a follow-up PR?

mcwhittemore:
@timiyay sorry, I went on vacation and dropped the ball here. Taking a detailed look now.

mcwhittemore:
Merged via #488

PR #478, opened 2016-08-14
New test structure for interaction events

### LINKED ISSUE #473: Split up interaction_events.test.js into smaller independent units`
Right now any modifications to `interaction_events.test.js` can be a huge pain, because most of the tests in there rely on continuous shared state — so if you modify something towards the top, you'll need to modify stuff below. This was a mistake I made.

Those tests should be split up into smaller, independent units that can be modified without requiring modification of other unrelated tests.

davidtheclark:
Closes #473.

The goal here, as described in the referenced issue, is to split up interaction event tests so each is independent of the others. This will make them much easier to work with, add to, modify, etc.

Here's what I've done: I made a new `createTestMap` test utility, which mounts a map and provides a `testMap` object with all kinds of functions for interacting with and deriving information from the map and its features. It's kind of like the Page Object pattern in UI testing, I think (?? from what I've read about it ...). I think it cleans things up nicely, allowing us to create and destroy an independent map for each test case without too much boilerplate.

I noticed that coverage went down slightly, though I tried to port over the equivalent of every test. I  can look more into this and maybe add more tests where I missed something. (I think the improved organization here also helps make it clear where there could be still other holes that would call for new test cases to be writtten at some point.) Additionally, there's still more cleaning up that could be done, of course, for total consistency of implementation across these tests, some that might be worth the trouble. 

But despite those two points that might hold up a merge, I wanted to get this PR up early because it's clearly a lot of change and I'd like to see what you think of the pattern.

@mcwhittemore for review.


davidtheclark:
@mcwhittemore I think this is ready to merge if you like the changes. Please have a look sometime and let me know if you think we should change anything.

mcwhittemore:
@davidtheclark this looks like a good idea to me. Nice linting rules too :)

mcwhittemore:
As there is no functionality change, I don't think we need to cut a new version. Merging.

PR #481, opened 2016-08-25
Fix IE11 crash

timiyay:
* Avoiding argument-less calls to classList.remove and add

### DIFF
--- a/src/ui.js
+++ b/src/ui.js
@@ -39,8 +39,13 @@ module.exports = function(ctx) {
       }
     });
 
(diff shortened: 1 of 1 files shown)


timiyay:
I can confirm my PR fixes the issue in my app, with no observable side effects.

Due to the miniscule nature of the change, and time constraints, I didn't add a test.

If you'd like coverage for this, I would intend to use Sinon to spy on `classList.remove` and `classList.add` and ensure they don't get called.

mcwhittemore:
@timiyay this looks good to me.

mcwhittemore:
@timiyay this has been published via 0.11.15

Issue #484, opened 2016-08-29
MultiLineString error


ilyamilosevic:
Trying to move vertex of MultiLineString geometry I get an error: `Uncaught TypeError: Cannot read property 'getCoordinate' of undefined`. Example code:

```js
mapboxgl.accessToken = 'YOUR_ACCESS_TOKEN';

var map = new mapboxgl.Map({
    container: 'map',
    style: 'mapbox://styles/mapbox/streets-v8',
    center: [30.34859077404613, 59.90705669106734],
    zoom: 12
});

var Draw = mapboxgl.Draw();

map.addControl(Draw);

map.on('load', function() {
    Draw.add({
        "type": "FeatureCollection",
        "features": [{
            "type": "Feature",
            "properties": {},
            "geometry": {
                "type": "MultiLineString",
                "coordinates": [
                    [
                        [
                            30.321750640869137,
                            59.917774138827866
                        ],
                        [
                            30.323638916015625,
                            59.88979807258723
                        ],
                        [
                            30.372734069824222,
                            59.89065922608917
                        ],
                        [
                            30.365352630615234,
                            59.92302240466373
                        ]
                    ],
                    [
                        [
                            30.372047424316403,
                            59.894189722105466
                        ],
                        [
                            30.38852691650391,
                            59.90322952572074
                        ],
                        [
                            30.386810302734375,
                            59.91278321461452
                        ]
                    ]
                ]
            }
    }]});
});
```

Environment:
- mapbox-gl 0.21.0,
- mapbox-gl-draw 0.11.13,
- Chrome 47.0.2526.80 (64-bit),
- Fedora 23.

I've opened a PR for this at #485.

PR #485, opened 2016-08-30
Check `lineBasePath` parameter

ilyamilosevic:
PR for issue #484.

### DIFF
--- a/src/lib/create_supplementary_points.js
+++ b/src/lib/create_supplementary_points.js
@@ -27,7 +27,7 @@ function createSupplementaryPoints(geojson, options = {}, basePath = null) {
     let firstPointString = '';
     let lastVertex = null;
     line.forEach((point, pointIndex) => {
-      const pointPath = (lineBasePath !== undefined) ? `${lineBasePath}.${pointIndex}` : String(pointIndex);
+      const pointPath = (lineBasePath != undefined) ? `${lineBasePath}.${pointIndex}` : String(pointIndex);
       const vertex = createVertex(featureId, point, pointPath, isSelectedPath(pointPath));
 
       // If we're creating midpoints, check if there was a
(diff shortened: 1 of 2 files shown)


ilyamilosevic:
O-o-ops... It's a bummer. Didn't run the tests.

mcwhittemore:
@ilyamilosevic thanks for this fix. We use it looks like this value can be null at times which is causing your problem.

- Expected: `coord_path: '0'`
- Found: `coord_path: 'null.0'`

ilyamilosevic:
@mcwhittemore, thanks you guys, too, for sharing this another good Mapbox project! Unfortunately, if I change comparison to `lineBasePath != undefined` (which, if I not mistaken, includes `null`) tests fail again with message: `# createSupplementaryPoints with MultiLineString, midpoints, selected coordinate not ok 159 should be equivalent`.

mcwhittemore:
@ilyamilosevic hmmmm... this could be a failure in our test suite... what is the `typeof lineBasePath`? Any chance that could work as a check for you?

ilyamilosevic:
@mcwhittemore, nope, not working. By the way, why for first LineString in MultiLineString geometry `coord_path` property of corresponding supplementary points is not two-dimensional? For example, on this line: https://github.com/mapbox/mapbox-gl-draw/blob/master/test/create_supplementary_points.test.js#L397

mcwhittemore:
Vertexes related to LineStrings should have a one-dimensional coord_path as this tells which Position the vertex relates to. For MultiLineStrings it should be two-dimensional (telling us which line and than which Position)

mcwhittemore:
@ilyamilosevic I've rebased and merged your via #485. Thanks!

ilyamilosevic:
@mcwhittemore Great! I've closed issue #484.

PR #486, opened 2016-08-30
Fix geojsonhint error filter

davidtheclark:
The latest version of geojsonhint uses `level: "message"` instead of `level: "warn"` — so I just switched this to a positive check for `level: "error"`.

@mcwhittemore for review. If you think this is good, let's release so we can take advantage downstream.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -50,7 +50,7 @@ module.exports = function(ctx) {
 
   api.add = function (geojson) {
     var featureCollection = normalize(geojson);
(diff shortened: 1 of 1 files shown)


mcwhittemore:
👍

davidtheclark:
Ugh this was a mistake. Turns out there's no `error` level either. There's *only* a `level: "message"` when it's *not* a layer. Will cut a new release.

PR #503, opened 2016-09-16
don't create a mapboxgl object on the window if its now there

### LINKED ISSUE #373: Draws adds mapbox-gl-js to the global scope even when complied with browserify
image

mcwhittemore:
Closes #373

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,5 +26,6 @@ var Draw = function(options) {
 
 module.exports = Draw;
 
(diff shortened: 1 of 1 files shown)


mcwhittemore:
The mapbox-gl-js plugin system is going to change soon. Closing this as we wait for that.

https://github.com/mapbox/mapbox-gl-draw/pull/525#issuecomment-258441677

PR #510, opened 2016-09-27
featuresAt returns unique features only

### LINKED ISSUE #509: Overlapping points at 0,0 obstruct dragging
If you create two points at 0,0 (I just plug `Draw.add({ type: "Feature", properties: {}, geometry: { type: "Point", coordinates: [0,0]}})` into the browser console when running the dev server), you'll have some challenges dragging them.

If I place the cursor over the center of the points, I can't drag. If I move it to the *edge* of the points, I can.

I tried creating two points at other coordinates, and the same problem did not recur.

davidtheclark:
Fixes #509.

The problem was that `queryRenderedFeatures` was returning multiple versions of the same feature, as it can do, and we were not filtering out those duplicates. With this change I can no longer reproduce the bug.

@mcwhittemore for review.

### DIFF
--- a/src/lib/features_at.js
+++ b/src/lib/features_at.js
@@ -1,6 +1,7 @@
 var sortFeatures = require('./sort_features');
 var mapEventToBoundingBox = require('./map_event_to_bounding_box');
 var Constants = require('../constants');
+var StringSet = require('./string_set');
 
 var META_TYPES = [
   Constants.meta.FEATURE,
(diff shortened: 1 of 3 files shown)


src/lib/features_at.js
```diff
@@ -24,5 +25,14 @@ module.exports = function(event, bbox, ctx) {
       return META_TYPES.indexOf(feature.properties.meta) !== -1;
     });
 
-  return sortFeatures(features);
+  var featureIds = new StringSet();
+  var uniqueFeatures = [];
+  features.forEach((feature) => {
+    const featureId = feature.id || feature.properties.id;
```
mcwhittemore: why `feature.properties.id`?

davidtheclark: That's where I was finding the ids in the features that Draw created. Fixing that to match the spec by putting the id directly on Feature could be another issue.

mcwhittemore: oh, this is in `features_at`. We should not use `feature.id`. While it should always be undefined the vector-tile-spec says it must be a number so we don't use it and thus there is no reason to check it. `feature.properties.id` is the right `id` in this case.

mcwhittemore:
@davidtheclark wanna drop `feature.id` and then merge and 📦.

PR #516, opened 2016-10-05
fix multifeature edit bug

ghoshkaj:


ghoshkaj:
@mcwhittemore for review

Issue #520, opened 2016-10-12
determine what is directly selected


jisaacks:
When switching to a **direct_select** mode you can select individual points of a polygon. However the `getSelected/getSelectedIds` still returns the entire polygon. Is there anyway to determine if the entire polygon or just some of it's points are selected?

I am building my own controls. When I call `draw.trash()` when in **direct_select** mode it only deletes points if points are selected. It does nothing if no points are selected, it will not delete the selected polygon.

What I would like to do is, if in **direct_select** mode and a polygon is selected entirely (not just some of it's points) and the user clicks trash, then delete the entire polygon (as opposed to do nothing) but there seems to be no way to make that distinction.

Or possibly hide the trash button when in **direct_select** mode unless a point is selected.

Either way I need to be able to make the distinction.

mcwhittemore:
@jisaacks - selected coordinates is managed by the direct_select mode rather than by Draw. I'm up for a PR that finds a good way to manage this in the store like we do with selected features.

`trash` in direct_select will always only deleted the selected coordinates. When this happen it will result in the feature being updated. `Draw.delete(id)` will delete a feature. You could do some sort of check to see if `trash` updated a feature and if it did not, delete the feature.

jisaacks:
@mcwhittemore I happy to work on a PR for this.

I think there are two possible solutions:

* Expose a function like `getSelectedCoordinates()` with an `draw.coordselectionchange` event

Or

* Make `trash()` return a boolean indicating if there were side effects
* Add a method like `trashable()` that indicates if `trash()` will produce side effects
* Add an event like `draw.trashablechange` that fires when the result of `trashable()` for the current mode changes.

I think I am in favor of the `trashable` solution because I think at the end of the day, that's what is hidden/cared about across modes.

Let me know what you think.

mcwhittemore:
So, we should keep `combineFeatures()` and `uncombineFeatures()` features in mind while we think about any solutions that change `trash`. All three of these functions are mode dependent. A way for the api to give developers a way to access what the mode/draw does itself.

Having `trash()`, `combineFeatures()` and `uncombineFeatures()` return false if they had no effect and return true if they had an effect seem like a good idea to me.

I'm not 100% sure what you mean with `trashable()`. Is this a function on `Draw` like `trash`?

All of these things are controls. Another option we've floated around is having a `draw.controls_changed` event. This event would fire any time a control becomes available or unavailable.

jisaacks:
[...]

PR #522, opened 2016-10-17
Add Draw.actionable events

jisaacks:
This is in response to discussion on #520 

ping @mcwhittemore


src/modes/simple_select.js
```diff
@@ -26,6 +27,32 @@ module.exports = function(ctx, options = {}) {
...
+      feature => feature instanceof MultiFeature
+    );
+
+    var combine = false;
+
+    if (selectedFeatures.length) {
+      var featureType = selectedFeatures[0].type.replace('Multi','');
+      selectedFeatures.forEach(feature => {
+        if(feature.type.replace('Multi','') === featureType) {
+          combine = true;
```
mcwhittemore: We should flip this logic to start as true and move to false if something that doesn't match the `featureType` is found.

src/lib/actionables.js
```diff
@@ -0,0 +1,17 @@
...
+    const nextActionables = cb();
+    if (nextActionables.uncombine === prevActionables.uncombine &&
+        nextActionables.combine === prevActionables.combine &&
+        nextActionables.trash === prevActionables.trash) {
+          return; // Nothing changed
+    }
+    ctx.map.fire(Constants.events.ACTIONABLE, nextActionables);
+    Object.assign(prevActionables, nextActionables);
+  }
+}
```
mcwhittemore: Hmmm... I bit worried about doing a noop check for some actionable changes and not for others. But I think this is ok for now. We can think a bit more about how to deduplicate these events a bit more later on.

jisaacks: That makes sense, so should I remove the deduping effort here until there is a more broad solution?

jisaacks:
@mcwhittemore Yep, I'll make the requested changes.

src/lib/actionables.js
mcwhittemore: +1

jisaacks:
@mcwhittemore I made the requested change. Also let me know how those tests look. I am sure there could be more tests for several scenarios but I think that is a good start.

mcwhittemore:
@jisaacks - this is looking good. Did you not want to add tests for the `direct_select` events as well?

Looks like this is conflicting with master. Can you rebase and I can do a final functional review and then we can merge this.

jisaacks:
@mcwhittemore, sure, if I don't get to it today I'll do it tomorrow. Thanks

jisaacks:
@mcwhittemore I added the requested tests and rebased. Thanks.

mcwhittemore:
@jisaacks - Thanks. This is great. I changed the structure of the event a bit so future values won't get overwritten if they conflict with mapbox-gl-js values. I also changed `combine` to `combineFeatures` and `uncombine` to `uncombieFeatures` to match the Draw functions.

Thanks for this! I'm going to go incorporate it into a project now :)

This has been published as `mapbox-gl-draw@0.13.0` on npm.

jisaacks:
Thanks @mcwhittemore, those changes you made make sense. I am going to be building something using this as well. If I encounter any issues I'll contribute back. :)

mcwhittemore:
Sounds great. Thanks!

PR #524, opened 2016-10-21
Add actionable events to trash methods for simple and direct select

### LINKED ISSUE #523: draw.actionable not firing after simple_select trash event
When in simple_select, and invoking `trash()` I should get a new actionable event indicated that trash is now false. This does not happen.

I believe the problem is because simple_select's trash method does not invoke `fireActionables()`. I did not add it there because I was under the impression that a call to `trash()` would always propagate to another `render()` which also calls `fireActionables()`. But that does not seem to be the case?

direct_select mode doesn't seem to be affected by this.

jisaacks:
Fixes #523


mcwhittemore:
Thanks. Merged this in and published as `mapbox-gl-draw@0.13.1`

PR #525, opened 2016-10-24
Export UMD modules

bensleveritt:
Updated the build commands to export UMD modules, which should be compatible with conventional usage, in addition to Webpack and non-module loading environments.


tmcw:
I think we should hold off on this change: we're narrowing in on a [plugin architecture for gl-js](https://github.com/mapbox/mapbox-gl-js/issues/1835) that will make the pattern of adding `Draw` to the `mapboxgl` object forbidden, and instead this module would be named `mapboxGLDraw` or some other independent variable.

bensleveritt:
That sounds like a good move. I suspect this change will still work after the change, but well worth holding off until you've settled on the new architecture.

PR #529, opened 2016-11-01
'line_string' Option Correction in API Docs

mrblueblue:
`line` is not a valid option -- it seems that the doc is actually referring to `line_string`


davidtheclark:
Thanks @mrblueblue!

PR #531, opened 2016-11-03
Add touch interaction

ChrisBellew:
Thanks for the great library guys! It's really great and has given my team a lot of features that are really useful.

One thing we needed to support our customers is touch interaction. So I have added the ability to draw, edit and move with tapping and touch-moving.

In high level terms I have:
- Added optional touch behavior.
- Added touch handlers that handle touch events and stop the events from propagating through to the mouse handlers.
- Added a more generous tolerance for feature collision detection when the event is detected as being a touch. This tolerance is configurable.
- Tests for the above functionality.

This seems like a fairly useful set of features for other developers so I'd like to push them into master. Please let me know if there is anything I can do to help.

Thanks!
Chris


mcwhittemore:
@ChrisBellew this is great! Thanks for all your work!

From a functionality POV I think we're pretty close. Here are a few problems I've seen playing with it on my phone.

- The first position of a line_string or a polygon isn't displayed until the second position is active. [A render option if these features have only one Position should solve this.](https://github.com/mapbox/mapbox-gl-draw/blob/bbfd25cba32761207f01e8953f86891f076bcb69/src/modes/draw_polygon.js#L95)
- Closing a shape by double tapping doesn't work as expected. If I triple tap it seems to work but creates an extra position. My guess is that something is getting confused about what the last Position created is.
- Selecting multiple features is impossible (I think this can wait till a later PR)

API.md
```diff
@@ -23,8 +23,10 @@ option | values | function
 --- | --- | ---
 drawing | boolean | The ability to draw and delete features - default: `true`
 keybindings | boolean | Keyboard shortcuts for drawing - default: `true`
+touchEnabled | boolean | If true, enables touch interaction - default: `true`
 boxSelect | boolean | If true, shift + click to features. If false, click + select zooms to area - default: `true`
 clickBuffer | number | On click, include features beyond the coordinates of the click by clickBuffer value all directions - default: `2`
+touchBuffer | number | On touch, include features beyond the coordinates of the touch by touchBuffer value all directions - default: `25`
```
mcwhittemore: The values for `clickBuffer` and `touchBuffer` are pretty different. My guess is this is because a mouse click is more precise than a finger tap, but we might want to explain these numbers a bit more to help people understand how to optimize.

src/events.js
```diff
@@ -70,6 +81,51 @@ module.exports = function(ctx) {
...
+      return;
+    }
+
+    touchStartInfo = {
+      time: new Date().getTime(),
+      point: event.point
+    };
+    var target = featuresAt.touch(event, null, ctx)[0];
+    event.featureTarget = target;
+    currentMode.touchstart(event);
```
mcwhittemore: @ChrisBellew what are your thoughts on mapping touch events and map events to the same event handlers? Is there a time when a `touchstart` event is not the same as a `mousedown` event?

One place I could see this as being different is in multi-select. A `tap` event and a `click` event might want to work differently to enable touch-based multi-select. That said, I think this can still be down my mapping the events to the same handler. Rather than passing the raw event to the handler, we'd need to pass our the interpreted event.

src/lib/is_tap.js
```diff
@@ -0,0 +1,16 @@
+const euclideanDistance = require('./euclidean_distance');
+
+const TOLERANCE = 25;
+const INTERVAL = 250;
+
+module.exports = function isTap(start, end, options = {}) {
```
mcwhittemore: @ChrisBellew - what are your thoughts on using `is_click` to do this check? The code is very similar here. Also, any notes on why we're using `250` and `25` for the interval and tolerance here? The `is_click` values is from experience from the iD editor.

ChrisBellew:
Thanks @mcwhittemore. I'll definitely look into the feedback for the three
pull requests, but it will have to be next week.

src/lib/is_tap.js
ChrisBellew: I thought about this too. 

Just calling is_click inside a touch handler might seem confusing to the reader but perhaps renaming it to something that an is_click and is_tap could both call would work. But I suppose the real reason I split them is because I felt that touch events do not need to gross/fine tolerance mechanism because maybe a finger is less likely to move once it has it the screen? I don't have any experience in this so it was pure hypothesis, and perhaps I am spouting ignorance.

What do you think?

src/events.js
ChrisBellew: I wrestled with this in my head. It's obviously a DRY-explicitness tradeoff. My personal preference is to make things explicit when the interaction mechanism is fairly different in order to make changes easier later.

But that's just my opinion. If you think it would be better to have handlers with one code path then I'll go ahead and change them?

API.md
ChrisBellew: Added a bit more explanation, but I'm afraid I can't back it up with science. It just seemed to feel right on the devices i tried.

src/lib/is_tap.js
mcwhittemore: > the real reason I split them is because I felt that touch events do not need to gross/fine tolerance mechanism because maybe a finger is less likely to move once it has it the screen

This makes sense to me.

> calling is_click inside a touch handler might seem confusing to the reader but perhaps renaming it to something that an is_click and is_tap could both call would work

What about [checkInteraction](https://gist.github.com/mcwhittemore/74143b310bb70e58f326f82500707b90) which exports a `isClick` and a `isTap` function.

ChrisBellew: Thanks! I'll use that

ChrisBellew:
> - The first position of a line_string or a polygon isn't displayed until the second position is active. A render option if these features have only one Position should solve this.
> - Closing a shape by double tapping doesn't work as expected. If I triple tap it seems to work but creates an extra position. My guess is that something is getting confused about what the last Position created is.

I've spent a while looking into these two issues and it looks like they are both related to the way vertices are added in draw_polygon mode (possibly draw_string_mode as well but I haven't checked).

When starting to draw a new polygon with the mouse, the first `click` sets the position of the first coordinate and then when the next `mousemove` is fired, another coordinate is added. [In Chrome](http://stackoverflow.com/questions/14538743/what-to-do-if-mousemove-and-click-events-fire-simultaneously) this happens simultaneously. However, mapbox will not actually draw anything unless the coordinates that make up the polygon or line string are slightly apart from each other. When using a mouse all it takes is a tiny movement of the mouse to make this true, and for a point or connecting line to appear.

When interacting by tapping, there is no `mousemove`, or even `touchmove` to cause this small delta to arise, and so nothing is actually rendered. This is the cause of the first issue.

As mentioned before, the coordinate is not added until the `mousemove` happens in the mouse example. This causes the rendering to be always "one vertex behind" when it is rendering the polygon and its vertices. This is the cause of the second issue because it does not create the last vertex so there is nothing to `tap` on.

I've no doubt there is a good reason why it works this way, but it probably needs to be changed slightly to support the touch use case. The solutions I can think of are:
- Add a small delta immediately to the lat lng of the previous point and add that as another point. This should trick mapbox into actually rendering something.
- That's all I got :)

Any ideas?

mcwhittemore:
Right now `draw_polygon`'s render function ignores anything with less than 2 vertices. This is done so the act of moving a mouse before a vertex is placed doesn't force a bunch of rerenders. It also looks nicer (I think) than dragging a point. Once the first click happens we start to render and live update. This is done by watching the mouse move, but it could be done in other ways. Maybe the render function should watch some mode wide state variable?

https://github.com/mapbox/mapbox-gl-draw/blob/cdbb15f65b94b2853171108f4c876e3220cf7981/src/modes/draw_polygon.js#L86-L94

ChrisBellew:
Thanks for all your feedback.

I have fixed the two issues mentioned most recently by using the `currentVertexPosition` to determine what type of geometry to render in draw_polygon. `currentVertexPosition` is incremented on click and tap, and has no relation to lazy creation of vertices on `mousemove`, so it is immune to the differences between mouse and touch. Also, whenever `currentVertexPosition` is 1, then I render a Point, which gives the user feedback when placing the first vertex by tapping.

ChrisBellew:
Also, I've noticed that circleci doesn't appear to be running. As soon as I push changes it says all checks have passed...and I'm sure previously this used to take a minute or so? The tests are passing on my end so it should pass anyway, but it seems too quick to be true.

mcwhittemore:
@ChrisBellew - This is looking great. Let's get it updated with master and merged!

PS: I double checked circle and it looks like its run the tests.

ChrisBellew:
Updated. Thanks :)

mcwhittemore:
Closing and reopeing to see whats going on with circle...

mcwhittemore:
meh, circle has died on this PR...

ChrisBellew:
Hey @mcwhittemore, would me submitting a new PR help?

davidtheclark:
@ChrisBellew Yes: if you update your branch, it should use Travis instead of Circle — and CI should work.

PR #532, opened 2016-11-03
Set feature property

ChrisBellew:
I have a use case where i want to show the area of a polygon in its center. With Mapbox I can configure a style to have a point layer which labels the polygon with a feature property.

As such, I have added a setter to the API to allow outside code to set an arbitrary feature property, which can then be styled by Mapbox.

This could also be useful for holding state specific to a feature inside the feature itself and saves having to write an external source of truth and keeping it in sync.

Thanks, and please let me know if there's anything I can do in relation to this PR.
Chris :)

### DIFF
--- a/src/feature_types/feature.js
+++ b/src/feature_types/feature.js
@@ -26,6 +26,10 @@ Feature.prototype.getCoordinates = function() {
   return JSON.parse(JSON.stringify(this.coordinates));
 };
 
+Feature.prototype.setProperty = function(name, value) {
+  this.properties[name] = value;
+}
+
 Feature.prototype.toGeoJSON = function() {
   return JSON.parse(JSON.stringify({
     id: this.id,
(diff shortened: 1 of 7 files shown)


src/feature_types/feature.js
```diff
@@ -39,15 +43,19 @@ Feature.prototype.toGeoJSON = function() {
...
 Feature.prototype.internal = function(mode) {
+  var properties = {
+    id: this.id,
+    meta: Constants.meta.FEATURE,
+    'meta:type': this.type,
+    active: Constants.activeStates.INACTIVE,
+    mode: mode
+  };
+  for (var name in this.properties) {
+    properties[name] = this.properties[name];
```
mcwhittemore: @ChrisBellew I want to stay away from possible clashes for properties. What do you think about having all user properties prefixed with `user_`. My concern with the prefix option is that it means the user provided properties will appear to change their name when exported from Draw.

mcwhittemore:
@ChrisBellew - I'm pretty sure I can get this up to date with the changes that have happened over the last two days so I'm going to start.

mcwhittemore:
@ChrisBellew this has been merged via 6ffd451dea47a7687f6408a895227dc5e967202c.

PR #533, opened 2016-11-03
Add ability to get the currently selected vertices

ChrisBellew:
I want to implement behavior in my calling application which allows the user to delete a point (vertex) or a feature.

If a point is selected, I want to show a "Delete Point" button.
If a feature is selected, I want to show a "Delete Feature" button.

In order to achieve this I need to know whether there is a point or a feature selected, or neither.

Currently, the API allows the calling code to get the selected features, so I have added a function to also expose the selected points. Then I can know if any points or features are selected at any time.

Hope this is useful!
Thanks, Chris :)


src/modes/direct_select.js
```diff
@@ -73,6 +79,21 @@ module.exports = function(ctx, opts) {
...
   };
 
+  function pathsToCoordinates(paths) {
+    return paths.map(coord_path => feature.getCoordinate(coord_path));
+  }
+
+  function coordinatesToPoints(coordinates) {
+    return coordinates.map(coords => ({
+      type: Constants.geojsonTypes.FEATURE,
+      properties: {},
```
mcwhittemore: Let's store the coord_path and the feature_id here. This will let users be able to take this feature and take actions on the selected coordinates external to Draw if they please.

src/store.js
```diff
@@ -7,6 +7,8 @@ var Store = module.exports = function(ctx) {
   this._features = {};
   this._featureIds = new StringSet();
   this._selectedFeatureIds = new StringSet();
+  this._selectedPoints = [];
+  this._selectedPointsFeatureId = null;
```
mcwhittemore: I think we can drop this if we store the feature id as part of the selected coordinate properties

src/store.js
```diff
@@ -212,6 +217,32 @@ Store.prototype.setSelected = function(featureIds, options = {}) {
...
+ * Sets the store's points selection, clearing any prior values.
+ * Expects to be passed the feature id that the points belong to.
+ * @param {string} featureId
+ * @param {Array<Array<string>>} points
+ * @return {Store} this
+ */
+Store.prototype.setSelectedPoints = function(featureId, points) {
+  this._selectedPointsFeatureId = featureId;
+  this._selectedPoints = points;
+  this._emitSelectionChange = true;
```
mcwhittemore: Lets make sure to update what is getting emitted as well [here](https://github.com/mapbox/mapbox-gl-draw/blob/bbfd25cba32761207f01e8953f86891f076bcb69/src/render.js#L58)

src/store.js
```diff
@@ -228,10 +259,18 @@ Store.prototype.getSelected = function() {
 };
 
 /**
+ * Returns selected points in the currently selected feature.
+ * @return {Array<Object>} Selected points.
+ */
+Store.prototype.getSelectedPoints = function() {
+  return this._selectedPoints;
```
mcwhittemore: This is a mutable interface. While we don't currently have anything that depends on these values, we should still protect this data. To do this, we can hold the 'raw data' in store and then convert it to features in the api.js function.

I think we'd want to store to hold an object that looks like this.

```js
[{feature_id: '123', coord_path: '1.1.1', coordinates: [0,0]}}
```

@ChrisBellew does this seem reasonable? Is there something I'm over looking here?

src/modes/direct_select.js
mcwhittemore: As said on the comment I left in `store.js` I think converting this data into features should be done in `api.js`. This is to keep with how `store.js` currently works (it doesn't store user presented GeoJSON) and to keep the Draw API interface immutable.

ChrisBellew:
Thanks for the review @mcwhittemore and sorry for the delay...had to wait for a task to land in a sprint, you know? :)

I have made the changes and you recommended...hopefully I haven't misunderstood anything. Would you be able to review them for me please?

Thanks
Chris

src/store.js
```diff
@@ -169,6 +170,7 @@ Store.prototype.deselect = function(featureIds, options = {}) {
     if (!this._selectedFeatureIds.has(id)) return;
     this._selectedFeatureIds.delete(id);
     this._changedFeatureIds.add(id);
+    this._selectedCoordinates = this._selectedCoordinates.filter(point => point.feature_id !== id);
```
mcwhittemore: This will filter out all points that don't match all selected features. In our current use case, only one feature is selected when coordinates are selected, but if we change that this will create unexpected behaviors.

Let's file a clean up issue for this as its ok for now.

test/interaction_events.test.js
```diff
@@ -84,7 +84,8 @@ function runTests() {
       });
 
       firedWith(t, 'draw.selectionchange', {
-        features: [pointA]
+        features: [pointA],
+        points: []
```
mcwhittemore: @ChrisBellew it would be good to have a test that confirms that points are being emitted. We can file this as a follow up task if you want. The code for interaction tests is a bit wonky.

src/render.js
```diff
@@ -55,7 +55,17 @@ module.exports = function render() {
 
   if (store._emitSelectionChange) {
```
mcwhittemore: When deselecting a point by clicking on the feature in `direct_select` mode the `draw.selectionchange` does not emit. This action keeps the user in direct_select. If I click off of the feature (which moves me to `simple_select`) the event is emitted.

It should be emitted both times.

src/modes/direct_select.js
```diff
@@ -158,6 +166,7 @@ module.exports = function(ctx, opts) {
     },
     stop: function() {
       doubleClickZoom.enable(ctx);
+      ctx.store.clearSelectedCoordinates();
```
mcwhittemore: We need this on https://github.com/mapbox/mapbox-gl-draw/pull/533/files#diff-0210e4a9505172f8c13cd7a2cb7b6c07R164 as well.

src/store.js
mcwhittemore: Filed: https://github.com/mapbox/mapbox-gl-draw/issues/561

test/interaction_events.test.js
mcwhittemore: Filed: https://github.com/mapbox/mapbox-gl-draw/issues/561

mcwhittemore:
@ChrisBellew - OK. This is really close. I found one bug which should be a simple fix.

Can you also update the API docs to explain the change to this event and the API changes. Once those are in this is ready to merge.

Thank you so much!

src/modes/direct_select.js
mcwhittemore: When deselecting a point by clicking on the feature in direct_select mode the draw.selectionchange does not emit. This action keeps the user in direct_select. If I click off of the feature (which moves me to simple_select) the event is emitted.

It should be emitted both times.

test/interaction_events.test.js
ChrisBellew: Added to this PR 248d307

src/modes/direct_select.js
ChrisBellew: Added to this PR 39958b2

ChrisBellew:
Thanks for the review. Hopefully I have adequately address your comments. I have fixed the two side issues you raised and if...

```I found one bug which should be a simple fix.```

...is in reference to the draw.selectionchange not being emitted in direct_select mode then I believe I have fixed this as well. 

Let me know if there is anything else :)

mcwhittemore:
@ChrisBellew - this is looking good. I found a pretty big but on the master branch when checking this. Once that is closed I'll get the next release out.

https://github.com/mapbox/mapbox-gl-draw/issues/563

PR #538, opened 2016-11-08
Fix syntax error

mike-marcacci:
This was breaking post-processing of a bundled file


mike-marcacci:
Never mind, I see that we are now supposed to import the pre-packaged `mapbox-gl-draw/dist/mapbox-gl-draw`, following the new mapbox conventions.

PR #540, opened 2016-11-10
Add back layers and force redraw after map style change

### LINKED ISSUE #512: Draw plugin breaks when map style changes
When I draw something on map and I change the map style, draw plugin breaks and I can't draw anything more. Draw plugin throws me this error:

`"The layer 'gl-draw-polygon-fill-inactive.cold' does not exist in the map's style and cannot be queried for features."`

Example:
http://codepen.io/Naimikan/pen/JRJOVN?editors=1111

I want to draw a polygon on style, save it, change style and continue drawing other polygons on other styles. I think I have to create style's layer on style change because style change won't do it by itself.

jisaacks:
Fixes #512 

This seems to fix it. Not sure it is the best solution, I had to attach `setup` to the `context` so I could re-invoke `setup.addLayers`.

I also has to first change mode to "static" before changing it back to what it previously was (forcing a re-render) otherwise the change mode is ignored.

If you want me to change anything let me know.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,8 @@ var Draw = function(options) {
   ctx.api = api;
 
   var setup = Setup(ctx);
+  ctx.setup = setup;
+
   api.addTo = setup.addTo;
   api.remove = setup.remove;
   api.types = Constants.types;
--- a/src/events.js
+++ b/src/events.js
@@ -104,6 +104,20 @@ module.exports = function(ctx) {
     ctx.store.changeZoom();
   };
 
+  events.data = function(event) {
+    if (event.dataType === 'style') {
+      const { setup, api, map, options } = ctx;
+      const hasLayers = !!options.styles.find(style => map.getLayer(style.id));
+      if (!hasLayers) {
+        setup.addLayers();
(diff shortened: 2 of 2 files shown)


mcwhittemore:
merged via 7511c0142e8cae58c5f57803b62440ffea6b7ee1

mcwhittemore:
@jisaacks - thanks for this.

I changed two things before merging this. First I had `setup` add itself to `ctx` and second I used `store.setDirty()` and `store.render()` to force the render so we could avoid changing modes.

jisaacks:
Sounds good, thanks!

FYI, browsing the updated code, it looks like there is a mistake, [adding the event, when it should be removing it](https://github.com/mapbox/mapbox-gl-draw/blob/master/src/events.js#L179). Sorry about that! I can add a PR if you want but it's trivial and probably not worth one if you want to just change the word from `on` to `off`. Let me know.

Thanks!

mcwhittemore:
Nice catch. Thanks! 

c230ca1e1f37647f50a032ec0c50f88af2dc9844

PR #550, opened 2016-11-17
Fixed typo in build command

bensleveritt:
Needed the `t` to fix the UMD export


mcwhittemore:
😭 - Thanks @bensleveritt!

Issue #559, opened 2016-11-29
Allow the user to delete the last point while drawing a line or polygon.


OmniEarthDev:
Most drawing tools allow the user to delete the last point while actively drawing a line or polygon.  

Currently the delete key and escape removes the entire line or polygon. 

Ideally, this new feature would just use the delete key to remove the last point.
If the delete key is repeatedly hit then each subsequent point would be removed until the whole drawing is removed.

This would then still leave the escape key to remove the entire drawing.

mcwhittemore:
I'm locking functionality for v1.0.0, but I want to note that I view this as a BUG and not as an enhancement.

timotgl:
@mcwhittemore Just to confirm, is the escape key supposed to remove the entire drawing or do you consider this part of the bug?

We're using the boundlessgeo-sdk and are having trouble re-enabling drawing mode after escape was pressed.

mcwhittemore:
@timotgl - `esc` should cancel drawing and delete the feature in progress. `delete` should delete the last coordinate added.

`re-enabling drawing` should be just a call to [draw.changeMode](https://github.com/mapbox/mapbox-gl-draw/blob/master/docs/API.md#changemodemode-string-options-object-draw).

adrice727:
@mcwhittemore Do you know if there is currently a workaround for this?  I'd like to be able to use the `delete` key to remove the node.  It seems that the only way around having `delete` remove the entire polygon is to provide a custom mode and override the `onTrash()` method, but there is no way to know what triggered `trash()`, so this isn't really helpful.

TannerPerrien:
Each mode is responsible for building the shape of choice and hence would also need to be responsible for choosing how to handle an "undo the last point" type action. With only a little work you can extend an existing mode and add your custom flare.

Have a look at the current polygon mode:

```js
DrawPolygon.onTrash = function(state) {
  this.deleteFeature([state.polygon.id], { silent: true });
  this.changeMode(Constants.modes.SIMPLE_SELECT);
};
```

`onTrash` is called when the delete button is clicked or when `draw.trash()` is called. You could modify the modes of your choosing to instead remove the last placed point. For example:

```js
DrawPolygon.onTrash = function onTrash(state) {
  if (state.currentVertexPosition <= 1) {
    this.deleteFeature([state.polygon.id], { silent: true });
    this.changeMode(Constants.modes.SIMPLE_SELECT);
  } else {
    const ring = state.polygon.coordinates[0];
[...]

PR #569, opened 2016-12-18
#559  Allow the user to delete the last point while drawing a line or polygon.

ravraponi:
Thanks for creating this drawing plugin, it has been very useful and has saved me a lot of time.

I have added the following:

Changed the delete key behavior on drawing a line or polygon to:
Delete key removes the last point.
When "delete" is repeatedly hit, then each subsequent point is removed until the whole drawing is deleted.

Escape key will still remove the entire drawing.

Changed tests for the above functionality.


mcwhittemore:
@ravraponi - I'm not sure why the tests did not run here. Can you push a new line commit or something and see if circle picks this up?

ravraponi:
@mcwhittemore - not seeing any change with circle.  Any ideas?

mcwhittemore:
@ravraponi - I'm not seeing anything either. Maybe something changed in how circles handle external contributors tests? I'll look into this more later today.

davidtheclark:
@ravraponi Thanks for the contribution! Tests are passing now, after we switched to Travis.

I like the idea of the interaction. Although it's a breaking change, the break is very minor — this new interaction fits right into the existing ones.

When I tried this out, I found a few bugs, though.
When I draw a few vertices then, don't move my mouse after creating the last one, then hit Backspace, the line segment is no longer stretched out to my cursor — *until* I move, at which point the segment changes suddenly and strangely. The same thing happens with polygons. (This seems to be tied up with the fact that the new vertex does not change its appearance and behavior until I move my mouse away from it.)

jan-23-2017 07-48-55
jan-23-2017 07-52-07

By creating a 3-vertex polygon then hitting Delete, I can break Draw. It seems I end up sending a bad Feature to the map.

jan-23-2017 07-51-33

Also, can you please add to and modify the documentation so it includes the new behavior?

ravraponi:
@davidtheclark 
I added documentation in the API for shortcut keys.  Not sure if this is what you meant.

mcwhittemore:
What's up with this PR? It seemed close a few weeks back.

ravraponi:
On my side - just waiting for a review

davidtheclark:
@ravraponi I'm still seeing weird behavior weird I don't move the mouse after drawing the latest vertex. It's changed, though. Now, if I draw vertexes and leave the mouse on the latest one, I have to hit Delete *twice* before the line changes. 

Another point: If I mess up on the *second* vertex of a line or *second through fourth* vertex of a polygon, Delete destroys the entire shape. More intuitive behavior, I think, would be to remove the shape from the store (because it's no longer valid) but keep the existing vertexes in place. Then I could, for example, draw a line with one vertex, then another, then delete the second and *immediately draw a new one* instead of starting over from the first.

The issue with the polygon error seems to be fixed!

mcwhittemore:
@ravraponi thanks for continuing to push on this. So great to see!

ravraponi:
@davidtheclark 
Polygon/Line now deletes to the last point and still allows immediately drawing of a new vertex.

As for the weird behavior when you don't move the mouse after drawing the latest vertex and then hitting delete -- 
The seemly double delete is actually because of adding in the vertex for the cursor location. 
The cursor behavior was changed for your comment on Jan 23.  Try deleting once and then move the mouse and it will make more sense.

ravraponi:
@mcwhittemore @davidtheclark
This PR is ready for review

ravraponi:
@mcwhittemore @davidtheclark
This PR is ready for review.  Updated for line continuation an added test.  There is an odd Travis error that you will need to look at.

ravraponi:
bump

mcwhittemore:
@ravraponi I'm sorry having time to review this is so hard to find. @davidtheclark and I are both very bogged down with other projects right now and this PR requires a lot of fine combing to make sure it is ready. We're talking internally about how to fix this workflow. Thanks for your hardwork and patience.

mcwhittemore:
Modes have been greatly changed as of #671, so I'm going to close this PR for now. That said, this is a long reported bug and should be fixed in MapboxDraw rather than in a custom mode.

PR #574, opened 2017-01-04
Update mapbox-gl peerDependencies to 0.29.0

apkoponen:
The peerDependencies do not include the latest Mapbox GL JS version, which causes an annoying nag when using npm.


davidtheclark:
Thanks @apkoponen. 

@mcwhittemore any objection to me merging and releasing?

mcwhittemore:
This should be good to go. Looks like the tests are still not running... :(

davidtheclark:
@apkoponen Released this in 0.16.1.

PR #592, opened 2017-02-14
fixed typo in example

aatdark:
your styling example does not show points because both rules filter for active = true.
imho the second one should be false.


mcwhittemore:
This looks good

Issue #597, opened 2017-02-22
getSelectedPoints should be getSelectedVertices?


davidtheclark:
It seems to me that `getSelectedPoints` is a misnomer. The name suggests that I could have a polygon, two lines, and two points selected, and this function will return only the points. The true intention is to return the selected vertices of a directly selected feature, right?

mcwhittemore:
Yea... I missed this in the PR...

Bigger question, should we drop the use of `vertices` for the GeoJSON word `positions`?

davidtheclark:
> should we drop the use of vertices for the GeoJSON word positions

Sure!

mcwhittemore:
This has been well established. If we talk about doing a v2 of MapboxDraw, we should bring this back up.

PR #598, opened 2017-02-22
Documentation improvements

davidtheclark:
@mcwhittemore I tried to 

- Clarify some things
- Fix incorrect examples
- Use Flow-style annotations instead of the idiosyncratic annotations

I do not know what the last two tables are about so did not touch them. What *are* they for? We should clarify them.

### DIFF
--- a/API.md
+++ b/API.md
@@ -3,55 +3,60 @@
 To use Draw
 
 ```js
-var Draw = new MapboxDraw({ options });
-map.addControl(Draw);
+// Create a Mapbox GL JS map
+var map = new Map(mapOptions);
+
+// Create a Draw control
+var draw = new MapboxDraw(drawOptions);
+
+// Add the Draw control to your map
+map.addControl(draw);
 ```
 
-Draw only works after the map has loaded so it is wise to perform any interactions in the `load` event callback of mapbox-gl:
+**Draw only works after the Mapbox GL JS map has loaded**, so it is wise to perform any interactions after your maps `load` event:
 
 ```js
 map.on('load', function() {
-    Draw.add({ ... });
-    console.log(Draw.getAll());
-    ...
+  draw.add({ .. });
 });
 ```
 
 ## Options
 
-option | values | function
---- | --- | ---
-keybindings | boolean | Keyboard shortcuts for drawing - default: `true`
-touchEnabled | boolean | If true, enables touch interaction - default: `true`
-boxSelect | boolean | If true, shift + click to features. If false, click + select zooms to area - default: `true`
-clickBuffer | number | On click, include features beyond the coordinates of the click by clickBuffer value all directions - default: `2`
-touchBuffer | number | On touch, include features beyond the coordinates of the touch by touchBuffer value all directions - default: `25`
-displayControlsDefault | boolean | Sets default value for the control keys in the control option - default `true`
-userProperties | boolean | If true, user properties will be present and prefixed with `user_` on the feature objects via styling. - default `false`
-controls | Object | Lets you hide or show individual controls. See `displayControlsDefault` for default. Available options are: `point`, `line_string`, `polygon`, `trash`, `combine_features` and `uncombine_features`.
-styles | Array | An array of style objects. By default draw provides a style for you. To override this see [Styling Draw](#styling-draw) further down.
+All of the following options are optional.
+
+- `keybindings`, boolean (default `true`): Whether or not to enable keyboard interactions for drawing.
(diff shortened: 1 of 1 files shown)


mcwhittemore:
The second to last table is trying to capture the values and attributes available when styling draw.

Personally, I haven't thought much about this table in a long time as we settled on a style for Draw and Studio so long ago, but I suspect that any new user who needs wants to make Draw match their brand needs this. Really, this would be better on a docs website than in a readme.

The last table is trying to capture the values and attributes that will be returned via the `queryRenderFeatures` mapbox-gl-js call. The thought is that if you are building on top of Draw you need to know how to connect gl-js's results back to Draw. This is likely more of a table for contributors and advanced users of Draw.

API.md
```diff
@@ -3,55 +3,60 @@
...
-controls | Object | Lets you hide or show individual controls. See `displayControlsDefault` for default. Available options are: `point`, `line_string`, `polygon`, `trash`, `combine_features` and `uncombine_features`.
-styles | Array | An array of style objects. By default draw provides a style for you. To override this see [Styling Draw](#styling-draw) further down.
+All of the following options are optional.
+
+- `keybindings`, boolean (default `true`): Whether or not to enable keyboard interactions for drawing.
+- `touchEnabled`, boolean (default `true`): Whether or not to enable touch interactions for drawing.
+- `boxSelect`, boolean (default `true`): Whether or not to enable box selection of features with `shift`+`click`+drag. If `false`, `shift`+`click`+drag zooms into an area.
+- `clickBuffer`, number (default: `2`): On click, select features or vertices beyond the coordinates of the click by the `clickBuffer` value in all directions.
+- `touchBuffer`, number (default: `25`): On touch, select features or vertices beyond the coordinates of the click by the `touchBuffer` value in all directions.
```
mcwhittemore: What are `2` and `25` in units? Pixels?

API.md
```diff
@@ -3,55 +3,60 @@
...
 
-The modes names are available as an enum at `Draw.modes`.
+The mode name strings are available as an enum at `Draw.modes`.
 
 ### `simple_select`
 
 `Draw.modes.SIMPLE_SELECT === 'simple_select'`
 
-Lets you select, delete and drag features.
+Lets you select, delete, and drag features.
```
mcwhittemore: meh, oxford.

API.md
```diff
@@ -3,55 +3,60 @@
...
 
-Draw will transition into `simple_select` mode every time a single feature has completed drawing.
+Draw is in `simple_select` mode by default, and will automatically transition into `simple_select` mode again every time the user finishes drawing a feature or exits `direct_select` mode.
 
 ### `direct_select`
 
 `Draw.modes.DIRECT_SELECT === 'direct_select'`
 
-Lets you select, delete and drag vertices.
+Lets you select, delete, and drag vertices.
```
mcwhittemore: This also lets you drag the selected feature now.

API.md
```diff
@@ -75,227 +80,250 @@ Lets you draw a Point feature.
...
 
-This method returns a FeatureCollection of all the selected features. If nothing is selected, it will return an empty FeatureCollection.
+Returns a FeatureCollection of all the features currently selected.
 
 ---
-###`.getSelectedPoints() -> [Object: GeoJSON]`
+### `getSelectedPoints(): FeatureCollection`
 
-This method returns all vertices currently in a selected state in an array of GeoJSON points.
+Returns a FeatureCollection of Points representing all the vertices currently selected.
```
mcwhittemore: So... this returns the selected vertices, not the selected points. I just realized how confusing this name could be. We should think about changing it. We should also look to see how this works in `simple_select` vs `direct_select` mode.

davidtheclark: Ha, I thought the same thing: https://github.com/mapbox/mapbox-gl-draw/issues/597

API.md
```diff
@@ -3,55 +3,60 @@
...
+
+// Create a Draw control
+var draw = new MapboxDraw(drawOptions);
+
+// Add the Draw control to your map
+map.addControl(draw);
 ```
 
-Draw only works after the map has loaded so it is wise to perform any interactions in the `load` event callback of mapbox-gl:
+**Draw only works after the Mapbox GL JS map has loaded**, so it is wise to perform any interactions after your maps `load` event:
```
davidtheclark: > Should we go so far as saying required rather than wise?

Yes.

> Should we make all the Draw functions throw errors if load isn't done?

I think that would be a good idea, yeah.

Issue #595, opened 2017-02-22
getSelected() should return shape currently being drawn (maybe?)


drewbo:
Another option would be adding `getActive` to return this shape

keyofj:
"draw.update" returns the polygon that is adjusted. Does this not return selected points for other draw types?

mcwhittemore:
This is breaking change as thus [a great candidate for custom mode](471fdfa89eb26d90fbe16791f5fc56a33b9026dc).

PR #599, opened 2017-02-22
Set selected id to rendered inprogress lines and polygons

drewbo:
Naive approach to #595

### DIFF
--- a/test/draw_polygon.test.js
+++ b/test/draw_polygon.test.js
@@ -699,3 +699,25 @@ test('draw_polygon touch interaction', t => {
     t.end();
   });
 });
+
+test('draw_polygon is selected while in progress', t => {
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+  const map = createMap({ container });
+  const Draw = new MapboxDraw();
+  map.addControl(Draw);
+  const afterNextRender = setupAfterNextRender(map);
+
+  map.on('load', () => {
+    Draw.changeMode('draw_polygon');
+    mouseClick(map, makeMouseEvent(1, 1));
+    mouseClick(map, makeMouseEvent(2, 2));
+    mouseClick(map, makeMouseEvent(3, 3));
+    afterNextRender(() => {
+      t.equal(Draw.getSelected().features.length, 1, 'is selected');
+      Draw.changeMode('simple_select');
+      t.equal(Draw.getSelected().features.length, 0, 'no longer selected');
+    });
+  });
+  t.end();
+});
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -93,6 +93,9 @@ module.exports = function(ctx) {
       if (geojson.geometry.coordinates.length < 2) return;
       geojson.properties.meta = Constants.meta.FEATURE;
 
+      // if we're rendering it; mark it as selected
+      ctx.store.setSelected(line.id);
+
       if (geojson.geometry.coordinates.length >= 3) {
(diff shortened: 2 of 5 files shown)


mcwhittemore:
I think this makes sense. Can you add a test to make sure the feature is unselected when someone exits the drawing mode? I'm pretty sure this should happen automatically, but I'd like a test to confirm.

drewbo:
Thanks @mcwhittemore; I had to have the test wait for `afterNextRender` since the selected updating happens in the mode `render` function. Let me know if that is unwieldy at all and needs refactoring

test/draw_polygon.test.js
```diff
@@ -699,3 +699,25 @@ test('draw_polygon touch interaction', t => {
...
+test('draw_polygon is selected while in progress', t => {
+  const container = document.createElement('div');
+  document.body.appendChild(container);
+  const map = createMap({ container });
+  const Draw = new MapboxDraw();
+  map.addControl(Draw);
+  const afterNextRender = setupAfterNextRender(map);
+
+  map.on('load', () => {
+    Draw.changeMode('draw_line_string');
```
mcwhittemore: This should be `draw_polygon`

drewbo: Good call, this is what I get for copy and pasting

drewbo:
I didn't add anything to points since once it is placed, it returns to `'simple_select'` with the point selected (not affected by my changes)

mcwhittemore:
This is great. Can you rebase and then we'll merge.

drewbo:
Rebased

mcwhittemore:
Sorry, @drewbo I meant to rebase this with master so it was up to date. I can do this if the conflicts are a problem.

drewbo:
Ah, my bad. Updated (and I guess squashed too)

drewbo:
@mcwhittemore does this (or should it) emit a `draw.selectionchange` event?

mcwhittemore:
1. Entering a drawing mode should emit a `draw.selectionchange` if something was selected before.
2. Before a shape is valid, `draw.selectionchange` should not be emitted.
3. Once a shape is valid, ... I'm not sure, I think this is up for debate. /cc @davidtheclark 
4. Leaving a draw mode to simple_select
  - with a new shape should result in `draw.selectionchange` if the event **was not** fired in `3`
  - without a new shape should result in `draw.selectionchange` if the event **was** fired in `3`

drewbo:
@mcwhittemore this should be good to go then (it doesn't make any changes to the current events model)

mcwhittemore:
@drewbo the `draw.selectionchange` event shouldn't be emitted until the `Polygon` is valid. This is working of `LineString`.

drewbo:
@mcwhittemore updated the `Polygon` logic to not be selected (and thus not emit the event) until after the 3rd coordinate is drawn (even if it isn't fully added yet)

mcwhittemore:
This is looking good.

We need to update the docs fo [draw.selectionchange](https://github.com/mapbox/mapbox-gl-draw/blob/master/API.md#drawselectionchange)

mcwhittemore:
I'm just noticing that this is a breaking change. :(

mcwhittemore:
I'd like to see this worked out as a custom mode. One hope I have for custom modes, is that we can use them to beta test functionality changes and, if one custom modes gains large popularity, we can switch up how Draw does this.

[Please look here for info on writing custom modes](https://github.com/mapbox/mapbox-gl-draw/blob/master/src/modes/README.md#creating-modes-for-mapbox-draw).

PR #603, opened 2017-02-28
fix add() with specified feature id example

kriscarle:
The example had the id inside the geometry object.


davidtheclark:
Thanks @kriscarle!

Issue #605, opened 2017-03-06
Allow continuation of lines


drewbo:
After a line is added, it can only be edited by moving the vertices or adding midpoints. It would be a nice feature to be able to continue adding vertices like `draw_line_string` but from one of the endpoints. Proposed API would be either:
- `Draw.changeMode('draw_line_string, { id: [id of feature to continue from] })`
- `Draw.changeMode('draw_line_string, { feature: [feature to continue from or add and continue from] })`

Open questions:
- How to specify which endpoint to continue from
- How to incorporate this in the UI (if at all)

mcwhittemore:
@drewbo I'm pumped about this feature!

`coord_path` is exposed on all Positions in a feature. This is a string of `.` separated array indexes and is used in the `direct_select` mode to let users drag one of these Positions.

I'd propose this API:

`Draw.changeMode('draw_line_string, { feature: 'id', from: 'coord_path' })`.

We might want to enforce `coord_path` to either be the `first` or `last` position.

We should make sure this works with `MultiLineStrings`.

drewbo:
@mcwhittemore I like the idea of enforcing `first`/`last` or `start`/`end` for the position. Thinking about it more, users will likely not know which end of the line they are looking to extend from without doing `.get(id)` so does it make sense to have them send a coordinate instead?

mcwhittemore:
`first`/`last` and `start`/`end` only works for LineStrings and while this PR is focused on LineStrings, we need to make sure this can work for Polygons later. `coord_path` solves this.

This isn't a hard thing to compute, but it is a hard thing to explain. I've been trying to think about how util function would make this easier but don't think it can. Either it would return some new list of coordinates for a feature which is overly verbose and still requires you to know how to get the `coord_paths` out of the list (think MultiPolygon vs LineString) or you will need to pass in references to the until function to define what you want back which is really the same thing as creating the coor_path yourself.

TL;DR: lets use coord_paths and figure out how to document them well? Maybe a drawing or something? I feel like this is something a picture would capture better than words.

drewbo:
How will Polygon "continuation" be different than the current Polygon editing? Does it detach a vertex and allow drawing from there? I'm not sure there is as convincing a use case for Polygons to deviate from their current editing mode.

mcwhittemore:
You make a great point... :man_facepalming: 

I was thinking about adding inner rings which won't require a `coord_path`.

mcwhittemore:
[...]

PR #606, opened 2017-03-07
[WIP] Feature/605 line continuation

drewbo:
Implements line continuation, towards #605:

- [x] Needs a test or two
- [x] Implements "continue from beginning" by mutating the original feature to reverse the coordinates array. I didn't feel great doing it but if you don't, and just set `currentVertexPosition` to 0, subsequent edits get really wacky (it tries to edit the existing vertices in turn). Thoughts on how terrible this is?
- [x] I could go either way with `from` requiring `start`/`end` vs. requiring an actual coordinate but [leaning toward the latter](https://github.com/mapbox/mapbox-gl-draw/issues/605#issuecomment-284810522)

### DIFF
--- a/API.md
+++ b/API.md
@@ -324,7 +324,7 @@ Changes Draw to another mode. Returns the draw instance for chaining.
 
 The `mode` argument must be one of the mode names described above and enumerated in `Draw.modes`.
 
-`simple_select` and `direct_select` modes accept an `options` object.
+`simple_select`, `direct_select`, and `draw_line_string` modes accept an `options` object.
 
 ```js
 // `simple_select` options
@@ -341,6 +341,15 @@ The `mode` argument must be one of the mode names described above and enumerated
   featureId: string
 }
 ```
+
+```js
+// `draw_line_string` options
+{
+  // The id of the LineString to continue drawing
+  featureId: string,
+  // The point to continue drawing from
+  from: Feature<Point>|Point|Array<number>
+}
 
 ---
 
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "unassertify": "^2.0.3"
   },
   "peerDependencies": {
-    "mapbox-gl": ">=0.27.0 <=0.31.0"
+    "mapbox-gl": ">=0.27.0 <=0.32.1"
   },
   "dependencies": {
     "@mapbox/geojson-area": "^0.2.1",
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -14,11 +14,19 @@ module.exports = function(ctx, opts) {
     if (!line) {
       throw new Error('Could not find a feature with the provided featureId');
     }
-    // const from = opts.from;
-    // if (!from) {
-    //   throw new Error('Please use the `from` property to indicate which coordinate to continue the line from');
-    // }
-    currentVertexPosition = 0 //line.coordinates.length;
+    const from = opts.from;
+    if (!from) {
+      throw new Error('Please use the `from` property to indicate which coordinate to continue the line from');
+    }
+    if (from === 'end') {
+      currentVertexPosition = line.coordinates.length;
+    } else if (from === 'start') {
+      line.coordinates.reverse();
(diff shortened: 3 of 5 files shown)


mcwhittemore:
> Implements "continue from beginning" by mutating the original feature to reverse the coordinates array. ... Thoughts on how terrible this is?

I'd say this is pretty bad. The problem is that a user will not notice that this happens, so while 90% of the time the order doesn't matter, there is going to be times when it does and those users are not going to know why the order has been broken.

To do this I think we need to use [line.addCoordinate](https://github.com/mapbox/mapbox-gl-draw/blob/master/src/feature_types/line_string.js#L13) to get the new position into the line rather than just setting the position via `line.updateCoordinate`.

> I could go either way with from requiring start/end vs. requiring an actual coordinate but leaning toward the latter

I imagine that we'll want to go from `direct_select` mode to `draw_line_string` mode once we have this functionality. I guess we could do this by adding `start` and `end` to the corresponding positions in the line string. This doesn't really work well for polygons though.

One problem with the `coord_path` option is that then you'd need to figure out which direction you are adding in. This might be trivial I guess.

drewbo:
@mcwhittemore updated the logic to not mutate the array. It requires tracking an internal variable "direction" within the mode which is a little inelegant but works out nicely. Let me know if this approach looks correct and I can write some tests and make any requested changes to the `from`:`start`/`end` API

drewbo:
@mcwhittemore also updated the `from` property to take coordinate. So for an existing line with id as `'a'` and coordinates `[0,0]` and `[0,1]` you can call `Draw.changeMode('draw_line_string', { featureId: 'a', from: [0,0] })` or `Draw.changeMode('draw_line_string', { featureId: 'a', from: [0,1] })`. I like this pattern better because it is very transparent from the users' perspective. Three different errors are thrown (no matching feature, invalid or missing `from` parameter, or `from` coordinate not at the start/end of the coordinate array)

mcwhittemore:
@drewbo - I'm getting an error trying to use this. Here is the error and the JSON of the feature I'm trying to edit.

image

```
{
  "id": "15957cf9ad4c342a818b4d69323edfc4",
  "type": "Feature",
  "properties": {},
  "geometry": {
    "coordinates": [
      [
        -67.675781249997,
        4.214943141401079
      ],
      [
        -60.29296874999898,
        -31.952162238018474
      ],
      [
        -25.136718749999744,
        -27.371767300513696
      ],
      [
        -17.050781249993236,
        6.315298538345601
      ]
    ],
    "type": "LineString"
  }
}
```

drewbo:
@mcwhittemore that looks like the "right" error (although it could definitely use some copy editing). For that string you should either send `[-67.675781249997,4.214943141401079]` or `[-17.050781249993236,6.315298538345601]` (the endpoint coordinates) as the `from` property

mcwhittemore:
Ah. Thanks. This is looking good. Let's get the tests and docs updated and merge!

drewbo:
@mcwhittemore added tests and updated docs. One tiny lingering oddity: the current line drawing mechanism uses `line.updateCoordinate` with an index just beyond the current array to add new coordinates (this likely occurs in the `mousemove` listener on desktop and on the next click on mobile). The `'backwards'` continuation that this new feature requires (traversing the coordinates array in reverse) means that we have to explicitly call `line.addCoordinate` on click to add another dummy coordinate which can then be updated on move. So the tests have to [check for one extra coordinate](https://github.com/mapbox/mapbox-gl-draw/pull/606/commits/97b6805bcb6a723f2935f08e325597f160e12195#diff-d2ca2412af668e11090c45aae155d421R656) when counting in that direction (although it isn't saved and doesn't cause issues? maybe on mobile?). Concerns?

mcwhittemore:
@drewbo - I think this is good. I double checked that the coordinates look right and I think we're good. I'm not 100% sure if this will work right on touch, but I don't have a good way to test this right now so I'm going to merge and we can bug fix if needed.

mcwhittemore:
Github isn't getting that this has been merged via the command line.

001a23e7544e60a3e06406c7488b1e384aed0801

PR #607, opened 2017-03-08
Update acceptable peer dependencies

drewbo:


package.json
```diff
@@ -63,7 +63,7 @@
     "unassertify": "^2.0.3"
   },
   "peerDependencies": {
-    "mapbox-gl": ">=0.27.0 <=0.31.0"
+    "mapbox-gl": ">=0.27.0 <=0.32.1"
```
mcwhittemore: Can we just say `>=0.27.0`?

@lucaswoj - is gl-js stable enough that we can be very sure there are no major breaking changes between now and 1.0.0? Keeping this semver range up to date is a bit tedious

lucaswoj: I wouldn't recommend that. If you look at the release notes you will see occasional minor breaking changes. Officially semver only suggests it is safe to auto-update within patch releases before v1.

PR #616, opened 2017-03-31
provide custom modes

primozs:
Could something like this be a start? #289 
I guess this would be more useful to somebody that would build this library himself.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,10 +10,10 @@ const setupDraw = function(options, api) {
     options: options
   };
 
-  api = setupAPI(ctx, api);
+  api = setupAPI(ctx, api, options);
   ctx.api = api;
 
-  const setup = runSetup(ctx);
+  const setup = runSetup(ctx, options);
 
   api.onAdd = setup.onAdd;
   api.onRemove = setup.onRemove;
--- a/src/api.js
+++ b/src/api.js
@@ -16,9 +16,12 @@ const featureTypes = {
   MultiPoint: require('./feature_types/multi_feature')
 };
 
-module.exports = function(ctx, api) {
+module.exports = function(ctx, api, options) {
 
-  api.modes = Constants.modes;
+  api.modes = Object.assign({}, Constants.modes, options.modesConstants);
+  if (options.disableDefaultModes) {
+    api.modes = options.modesConstants;
+  }
 
   api.getFeatureIdsAt = function(point) {
     const features = featuresAt.click({ point }, null, ctx);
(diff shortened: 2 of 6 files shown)


mcwhittemore:
@primozs - Thanks for this! I don't think this will negitivly effect one using this via a CDN.

I want to spin static mode out of gl-draw with this push. If you want to continue proving this change out, I'd recomend dropping static mode from the mode constants and finding a way to expose it to the debug mode.

primozs:
Yes, it would not negatively effect CDN version but someone with custom modes needs might want to reuse more code from the lib.

I don't know about static mode but I need to implement some custom modes to see where I am and what problems I might stumble upon.

primozs:
@mcwhittemore I have updated this commit with possibility to disable default modes. I also wired up defaultMode option that was already available. 
In general this works good for me.

mcwhittemore:
Thanks! Can you provide an example so I know if I'm testing this the way you expect it to be used?

Regards,
Matthew

primozs:
I am using it like this:
- "disableDefaultModes": false by default, all default draw modes are instantiated. When "disableDefaultModes" is true and no "defaultMode" it will trow an error with explanation. 
- "metaTypes" Specify and merge custom geojson meta types with default meta types constants. Like this "mousemove" can find custom geojson elements same as Vertex element.
-  "modes" object of key value. key is custom mode constant name and value is custom mode instance. when disableDefaultModes is false (default) custom modes are merged with default ones otherwise only custom modes are passed trough ctx
- modesConstants: same type of object as draw constants.js passed trough ctx for convenience 

```
this.draw = new MapboxDraw({
      disableDefaultModes: true, 
      defaultMode: cuDrawConstants.modes.CU_SIMPLE_SELECT,
      metaTypes: [
        cuDrawConstants.meta.OZONE,
        cuDrawConstants.meta.TASK_LINE
      ],
      modes: {
        [cuDrawConstants.modes.CU_STATIC]: cuStatic,
        [cuDrawConstants.modes.CU_SIMPLE_SELECT]: cuSimpleSelect,
        [cuDrawConstants.modes.CU_DIRECT_SELECT]: cuDirectSelect,
        [cuDrawConstants.modes.DRAW_CU_RULER]: cuDrawRuler,
        [cuDrawConstants.modes.DRAW_CU_POINT]: cuDrawPoint,
        [cuDrawConstants.modes.DRAW_CU_TASK]: cuDrawTask
      },
      modesConstants: cuDrawConstants.modes
    });

```

mcwhittemore:
@primozs - thanks for your first pass at this. This has been finished via #671.

PR #633, opened 2017-04-26
Don't cache selected points separately from other store features

### LINKED ISSUE #632: Output of draw.getSelectedPoints is inaccurate after dragging
**mapbox-gl-js version**: 0.36.0
**mapbox-gl-draw version**: 0.17.0

### Steps to Trigger Behavior

 1. Select a line, then select a point on that line
 2. Drag the point to a different location
 3. Call `draw.getSelectedPoints`

### Expected Behavior

`getSelectedPoints` should return the updated coordinates of the point that you just moved.

### Actual Behavior

`getSelectedPoints` returns the original, un-moved coordinates.

From a quick glance, it seems that `store.setSelectedCoordinates` is only getting called at the [start of the drag](https://github.com/mapbox/mapbox-gl-draw/blob/5c589b6deccfe19baab9078bca606c7e0b4252dc/src/modes/direct_select.js#L68). Once you stop dragging, the coordinates cached in `store._selectedCoordinates` are not updated with the new geometry values.

Given the same user behavior, calling `draw.getSelected` instead of `draw.getSelectedPoints` correctly returns the updated line.

dereklieu:
The goal of this PR is to close #632 

### Summary

In `direct_select` mode, entering the mode or clicking on a vertex saves the parent feature ID, the coordinate index, and the geometry of the selected vertex.

When a user drags the point, the point remains selected, but it's geometry value saved to the store remains the same. So calling `draw.getSelectedPoints` inside of a `draw.update` event handler will frequently return stale geometry.

To fix this, I make the `store._getSelectedCoordinates` method retrieve the coordinate geometry based on the store's (up-to-date) features, rather than the cached value in `store._selectedCoordinates`.

I also stop including coordinates in the `_selectedCoordinates` array at all, so that `direct_select` now only saves the parent feature ID and coordinate index of the selected vertex (or vertices).

### DIFF
--- a/src/modes/direct_select.js
+++ b/src/modes/direct_select.js
@@ -78,7 +78,7 @@ module.exports = function(ctx, opts) {
   };
 
   function pathsToCoordinates(featureId, paths) {
-    return paths.map(coord_path => { return { feature_id: featureId, coord_path, coordinates: feature.getCoordinate(coord_path) }; });
+    return paths.map(coord_path => { return { feature_id: featureId, coord_path }; });
   }
 
   const onFeature = function(e) {
--- a/src/store.js
+++ b/src/store.js
@@ -257,7 +257,13 @@ Store.prototype.getSelected = function() {
  * @return {Array<Object>} Selected coordinates.
  */
 Store.prototype.getSelectedCoordinates = function() {
-  return this._selectedCoordinates;
+  const selected = this._selectedCoordinates.map(coordinate => {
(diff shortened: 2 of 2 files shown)


mcwhittemore:
@dereklieu - Thanks for this. Can you add a test case that fails before this change is made?

dereklieu:
@mcwhittemore I opened a separate PR for that test in #634. If the test is adequate, I can close this PR and merge the fix into the test branch PR. Just let me know how you want to do this.

Issue #637, opened 2017-04-28
0.17.4 has empty dist js file


ChrisBellew:
Hello,

I have just `npm install @mapbox/mapbox-gl-draw@0.17.4` and have noticed that the dist/mapbox-gl-draw.js is an empty file. In `0.17.3` the file is 100kB. Is this expected?

Thanks
Chris

mcwhittemore:
Hi Chris,

Its not... though I really don't like how we currently build the dist folder. I think right now this happens on `npm publish` but it should happen on `npm post-install` I think. I don't use Draw this way so if you have thoughts on how to do this correctly and help us make sure it always happen, I'm all ears.

Matthew

danielthall:
Same issue here. Commenting to follow this thread.

z0d14c:
Same issue as well :\ trying to upgrade to get the new peer dependency for mapbox-gl so i can update versions for that and I ran into this...

andrewharvey:
Checking out the process for mapbox-gl-geocoder at https://github.com/mapbox/mapbox-gl-geocoder/blob/master/CONTRIBUTING.md#deploying which says to run `npm run prepublish` before `npm publish`.

Perhaps the solution here is adding an instruction at the start of https://github.com/mapbox/mapbox-gl-draw#publishing to run `npm run prepublish`?

mcwhittemore:
I've cut a new release `0.18.0` and the `dist/mapbox-gl-draw.js` in non-empty when I install it locally. Can someone else confirm and then we'll close this.

z0d14c:
@mcwhittemore can you publish it on npm?

mcwhittemore:
@z0d14c - `0.18.0` is [on npm](https://www.npmjs.com/package/@mapbox/mapbox-gl-draw).

z0d14c:
works for me @mcwhittemore

mcwhittemore:
Thanks @z0d14c!

PR #646, opened 2017-05-15
include npm run prepublish in publishing instructions

andrewharvey:
See https://github.com/mapbox/mapbox-gl-draw/issues/637#issuecomment-301435761 for background.

Aims to ensure #637 is fixed in the next release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ npm run test
 To github and npm
 
 ```
+npm run prepublish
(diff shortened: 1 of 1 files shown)


mcwhittemore:
@andrewharvey doesn't `npm publish` run `npm run prepublish`?

andrewharvey:
I guess it does per https://docs.npmjs.com/misc/scripts.

PR #648, opened 2017-05-25
Change package.json to point to compiled files instead of source files

walkerburgin:
As it stands right now we have to override this in Webpack in order to target ES5. Is there any reason not to point `package.json` at the compiled files?

```
{
    // ...
    resolve: {
        alias: {
            "@mapbox/mapbox-gl-draw": path.resolve(__dirname, "../node_modules/@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.js"),
        },
    },
    // ...
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "homepage": "https://github.com/mapbox/mapbox-gl-draw",
   "author": "mapbox",
   "license": "ISC",
-  "main": "index.js",
+  "main": "dist/mapbox-gl-draw.js",
   "browserify": {
     "transform": [
       "babelify"


mcwhittemore:
I'm naturally against pointing `main` at compiled code. My basic logic is that while one person might want to require the ES5 code, another might want ES2016. As the browsers are moving towards ES2016, I think exposing ES2016 code makes more general sense. That said, I'll admit that my reasoning isn't great and I am open to hearing some other viewpoints on why we should export the ES5 code.

Questions I've not researched

- What does exposing a compiled file due to total size?
- How would someone require the ES2016 code?
- What other ways could we help people setup their projects than this?

The `dist/` is really only here for the Mapbox CDN. I've considered dropping this folder a few times as we've had problems with `npm publish` not building the file.

steaks:
I just ran into an issue where I broke my product on IE because when converting from the CDN to bundle I unknowingly pointed to source rather than compiled code in `dist/`.

Personally I was confused because in the README I read `var MapboxDraw = require('@mapbox/mapbox-gl-draw');` is to be analogous to `<script src='https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-draw/v1.0.9/mapbox-gl-draw.js'></script>`.

My opinion is that this package should be distributed in a way that either is:

a) Consistent with https://github.com/mapbox/mapbox-gl-js (which points at `dist/` via https://github.com/mapbox/mapbox-gl-js/blob/master/package.json#L5)

or

b) Inconsistent with https://github.com/mapbox/mapbox-gl-js and explains how to point to `dist/`. Additionally, it'd be nice to call out that this package diverges from mapbox-gl-js and why it diverges.

Issue #610, opened 2017-03-14
Improve drawing support for Polygon and LineString in touch environments


FlorianBurgevin:
Hello,

I'm not able to find how to display all the polygon points when i'm drawing one. Only the vertex points are visible.

Does the feature is available?

mcwhittemore:
In the drawing modes, only the first and last points are shown. This is to help cut down on the number of the things we are updating while drawing. Since this is when users see the most change, we want to make sure it is as performant as possible.

FlorianBurgevin:
Ok I understand thanks.

So the issue is the first clicked point and the second are not displayed. Vertex starts at the third point?

mcwhittemore:
The vertices aren't shown until the Polygon is valid (which is after the third click). Once the third click happens, there should be a dot on the first the last place you clicked. These dots can be clicked to finish drawing the polygon.

FlorianBurgevin:
Yes that's it.

Is there any way to show the first two points during the drawing?

My problem is that on mobile, it's hard to understand after first click because nothing happened.
That's why I want to show first and second point before the third point is clicked. This is good after the third point is clicked.

mcwhittemore:
ah... mobile... that would be a problem.

@FlorianBurgevin I'm more than willing to rethink this a bit. Would you be up to running out the PR?

@ChrisBellew did the work on getting touch support added in, so we might want to bring him into this convo as well.

FlorianBurgevin:
I can have a look at the code. But maybe it could be interesting to talk about the good behavior.

Is that always to draw first two points? Or at least an option to style them like the vertices.

mcwhittemore:
> maybe it could be interesting to talk about the good behavior.

I agree we should talk about what we want here before diving in. I just also want to make sure that someone in the convo has the time and will to do this. A good convo will no follow up won't solve our problem.

> Is that always to draw first two points? Or at least an option to style them like the vertices.

So, I think we need to think about this in the context of `touch` and in the context of `mouse`.

What we have right now works pretty well for `mouse`. We should the line as you are drawing and turn this into polygon when there are enough Positions added to the shape. Once we can show a polygon, we also start to show the `start` and `end` positions in the polygon to help users end drawing by clicking on either the `start` or `end`.

Now, in a touch environment, this doesn't work well because there is no event being fired as the user moves their hand to the next location they are going to touch. Without this `moving` events we are unable to draw the line that lets the user know drawing has begun.

[...]

PR #649, opened 2017-05-30
#610 mobile ux improvements

z0d14c:
This PR adds an initial vertex when drawing a new line or polygon. Although more efforts could be made, this significantly improves UX by informing giving the user feedback on their input vs. none until they make 2 or 3 taps.

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -142,7 +142,7 @@ location.hash.replace(/^#/,'').split('/').reduce(function(args, val, i, hash){
 console.log('use `loadData(url)` to load in FeatureCollections to Draw via url');
 console.log('5.51MB Polygons - https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_50m_geography_regions_polys.geojson');
 
-console.log('4.99MB LineStirings - https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_10m_rivers_north_america.geojson');
+console.log('4.99MB LineStrings - https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_10m_rivers_north_america.geojson');
 
 console.log('4.03MB Points - https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_10m_admin_1_label_points.geojson');
 
--- a/src/modes/draw_line_string.js
+++ b/src/modes/draw_line_string.js
@@ -69,7 +69,15 @@ module.exports = function(ctx, opts) {
       });
 
       this.on('click', CommonSelectors.true, clickAnywhere);
-      this.on('tap', CommonSelectors.true, clickAnywhere);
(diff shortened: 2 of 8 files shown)


mcwhittemore:
@z0d14c +1 to adding the handles from the start.

Before this PR, we only showed handles when the polygon was valid. But since we are now showing them from the start, lets make sure they are always showing. 

- First tap looks great!
- Second tap: both ends of the line should have handles but only the first does
- third Tap: first and last point should have points, but none do.
- Forth Tap: 1st and 4th positions should have points but 1st and 3rd due.
- 5+: like the forth tap, the last handle is one position back from where it should be.

z0d14c:
@mcwhittemore I believe we're in a good spot now. The tests had to be modified a bit to account for the extra node that exists more often now. I think it's all functioning 💯 %

mcwhittemore:
On mobile, when drawing a polygon, when I place the second point I just see two handles rather than handles points and a line.

Otherwise, this is looking good.

andrewharvey:
Looking good to me, I can confirm seeing what @mcwhittemore mentioned already.

z0d14c:
@mcwhittemore @andrewharvey ive fixed the behavior up to your specs, but now the polygon integration test is failing. honestly i cant figure out what the problem is and i think it may be a "false" failure, with jsts.min.js:9 throwing a cryptic error. specifically this occurs on the drawGeometry call in the `Polygon integration` test. Would love any input or assistance with this; drawing seems to work fine in an actual browser environment on mobile or desktop chrome.

mcwhittemore:
@z0d14c - I'm looking at the tests now

mcwhittemore:
@z0d14c - I added some console logging to `test/utils/draw_geometry` to capture when in the drawing processes this error was thrown.

```
pre-click [ 0, 0 ] false
post-click
pre-click [ 30, 15 ] false
```

The error happens after the second click. Looking at your PR, you changed the number of coordinates used for different rendering types in a few places. My guess is that this is resulting in a polygon with two points being rendered in the tests. This is an invalid polygon and would cause this kind of error.

I hope this helps as a place to start. Sorry if working with the tests is kind of hard. :(. I can also take a look tomorrow.

z0d14c:
great feedback, i had figured out that it was failing on the 2nd point but i couldnt figure out why. thanks and ill try to get to this asap.

z0d14c:
@mcwhittemore @andrewharvey I think we're in a good place now.

mcwhittemore:
@andrewharvey - does this PR solve this problem? https://github.com/mapbox/mapbox-gl-draw/issues/650

andrewharvey:
@mcwhittemore It seems to be unrelated, ie. I notice #650 in both master and the latest version of this PR. So let's not let #650 hold this PR up.

z0d14c:
can we go ahead and merge?

z0d14c:
(I can't click the merge button)

mcwhittemore:
merged in 7a7313107d7d50b1ce664274823a7e6cb60c691e

PR #661, opened 2017-06-28
Fix touch events issue and the trash behaviour

sguignot:
Fix the behaviour of the trash button in "direct select" mode: it should just trash the selected feature as in "simple select" mode.
Fix touch behaviour: do not prevent default of touch events in "simple select" mode to let the click on any map features work (not only draw features).

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -83,7 +83,9 @@ module.exports = function(ctx) {
   events.touchstart = function(event) {
     // Prevent emulated mouse events because we will fully handle the touch here.
     // This does not stop the touch events from propogating to mapbox though.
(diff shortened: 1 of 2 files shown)


mcwhittemore:
@sguignot - Both of these bullet points are functionality changes, not bugs unless I'm not understanding your commits well. Because different users want different types of interactions we are not considering changes like these. The best option forward is [Custom Modes](https://github.com/mapbox/mapbox-gl-draw/issues/289).

mcwhittemore:
Thanks @sguignot. If you have thoughts on how Custom Modes could work, I'm eyeing to take another stab at this soon.

PR #670, opened 2017-07-10
add draw event example to API.md

### LINKED ISSUE #669: Add simple event listener examples to API.md
As a first-time user reading the docs, it was not clear to me how to implement an event listener.  I tried `draw.on('create')` before going back to read that they are namspaced as `draw.{event}` and are fired by the mapboxGL JS map object.

I think a simple snippet showing `map.on('draw.create', (d) => { ... });` would be helpful for noobs.  

Happy to PR this, just let me know if there is a preference for where it should go, format, etc.

andrewharvey:
closes #669


mcwhittemore:
@chriswhong - what are your thoughts? Is this enough?

chriswhong:
Yes, I think one example at the top of the section is fine.

PR #675, opened 2017-07-26
moved README.md for modes to separate docs folder

sclausen:
and adapt mentions of this file.

I created this pull request, because I have troubles compiling with `@angular/cli`, since it tries to interpret the README.md. Yes, I think you can fix this in @angular/cli, but despite that I also doubt, that a src folder is the right place for README.md files, so I created this pull request.


mcwhittemore:
I think this is a good idea, but we should move all the docs to a docs folder not just modes.

Also, lets use `docs/MODES.md` rather than creating the extra folder.

Thanks for the PR!

sclausen:
@mcwhittemore should I just move `/EXAMPLES.md` and `/API.md`, or some else?

mcwhittemore:
That should do it. Thanks.

mcwhittemore:
Thanks @sclausen

shongololo:
yes, this is a good change, webpack is trying to parse the markdown file!

shongololo:
Any chance this fix can be released in an incremental update?

mcwhittemore:
I just published [v1.0.1](https://github.com/mapbox/mapbox-gl-draw/blob/master/CHANGELOG.md#101) to expose this change.

PR #676, opened 2017-08-01
Fix API Doc example for getFeatureIdsAt

snkashis:
Object instead of 2 params.


mcwhittemore:
Thanks for this @snkashis. I've opened #677 as well, to help make sure these kinds of mistakes don't happen moving forward.

PR #685, opened 2017-09-11
Safely Remove Control

### LINKED ISSUE #572: Enabling and disabling draw seem hard to do transactionally safe
I have a single page application where I want to enable and disable draw based on navigational context. Entering some specific context may enable drawing while also updating some other state on the map (such as zoom, center, pitch etc.). As it seems, enabling and adding draw (and its inherent dependencies) may be queued _after_ such other operations, which makes it hard to determine when draw has actually been successfully enabled (i.e. controls, sources and layers good to go).

Since it's hard to tell when draw has been successfully enabled it's also hard to determine when/if one can safely remove draw without entering a degenerate state.

If you go to [this](http://codepen.io/anon/pen/BQXLmp) codepen you will have a setup where draw has not yet been enabled. If you just enable/disable draw by clicking on the add/remove buttons it works well, but if you have draw disabled and click the fly button and then try to enable draw it will add the controls, but the necessary layers will not be added until the fly is finished (which takes a while!). If you click on remove at this point you will get exceptions and enter a state hard to recover from (controls are still there but there are no sources/layers on the map).

If you think this scenario is extreme/unlikely you may actually provoke the same anomaly just by dragging the map and hit add/remove before the panning has finished. It seems that any current operations being performed by the map queues up the enabling of draw.

As you probably understand this behaviour makes it really hard to add/remove draw safely in a single page application where other map interactions are carried out simultaneously. Sometimes it works, sometimes it doesn't.

What I believe is required is either some kind of "promise" that tells when draw actually has been "completely" enabled, meaning that it's safe to remove (map.on("draw.load")?). Or that the removeControl operations can handle a situation where draw has not yet been completely enabled and still remove whatever is there.

gpbmike:
Removing the draw control before the map is loaded throws errors related to missing layers.

This PR introduces a two part solution:

1. When the control is removed, stop waiting for map to load.
1. Check for source/layer presence before attempting to remove them.

fixes #572 

I would love feedback.

### DIFF
--- a/src/setup.js
+++ b/src/setup.js
@@ -6,9 +6,14 @@ const Constants = require('./constants');
 module.exports = function(ctx) {
 
   let controlContainer = null;
+  let mapLoadedInterval = null;
 
   const setup = {
     onRemove: function() {
+      // Stop connect attempt in the event that control is removed before map is loaded
+      ctx.map.off('load', setup.connect);
+      clearInterval(mapLoadedInterval);
+
       setup.removeLayers();
       ctx.ui.removeButtons();
       ctx.events.removeEventListeners();
@@ -20,6 +25,12 @@ module.exports = function(ctx) {
       controlContainer = null;
 
       return this;
+    },
+    connect: function() {
+      ctx.map.off('load', setup.connect);
+      clearInterval(mapLoadedInterval);
+      setup.addLayers();
+      ctx.events.addEventListeners();
     },
     onAdd: function(map) {
       ctx.map = map;
@@ -38,20 +49,11 @@ module.exports = function(ctx) {
         map.dragPan.enable();
       }
 
-      let intervalId = null;
-
-      const connect = () => {
-        map.off('load', connect);
-        clearInterval(intervalId);
-        setup.addLayers();
-        ctx.events.addEventListeners();
-      };
-
       if (map.loaded()) {
-        connect();
+        setup.connect();
       } else {
-        map.on('load', connect);
-        intervalId = setInterval(() => { if (map.loaded()) connect(); }, 16);
+        map.on('load', setup.connect);
+        mapLoadedInterval = setInterval(() => { if (map.loaded()) setup.connect(); }, 16);
       }
 
       ctx.events.start();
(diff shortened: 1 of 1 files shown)


mcwhittemore:
@gpbmike - Thanks for this. I think this solves the bug when adding and remove Draw. I've tested the fiddle from #572 to see how adding polygons reacts when in a non set mode. There don't seem to be any errors and the data isn't added, so I think we're ok there.

Can you add a test that confirms that a if `map.loaded()` returns false `setup.connect` is not called and that then, when `map.loaded()` returns true `setup.connect` is called.

Also, can you add a test that confirms `removeLayers` does not error without the Draw having been fully added to the map.

gpbmike:
@mcwhittemore sure thing. Is there an existing test file that would make sense to add to or should I create a new one?

mcwhittemore:
> Is there an existing test file that would make sense to add to or should I create a new one?

Whatever is easiest for you works for me.

gpbmike:
For some reason I'm not able to run the test suite. It just hangs here after using `npm test` or `yarn test`. Any suggestions?

```
TAP version 13
# Draw.getFeatureIdsAt
```

mcwhittemore:
What version of node are you using?

gpbmike:
> What version of node are you using?

`v8.4.0`

mcwhittemore:
Travis is running `v6.x.x` - can you try in that range?

gpbmike:
No dice. Still hanging.

```
mike at 3DR-000316 in ~
$ nvm use 6
Now using node v6.11.3 (npm v3.10.10)

mike at 3DR-000316 in ~
$ which node
/Users/mike/.nvm/versions/node/v6.11.3/bin/node

mike at 3DR-000316 in ~
$ cd oss/mapbox-gl-draw/

mike at 3DR-000316 in ~/oss/mapbox-gl-draw on master [!?]
$ npm --version
3.10.10

mike at 3DR-000316 in ~/oss/mapbox-gl-draw on master [!?]
$ npm test

> @mapbox/mapbox-gl-draw@1.0.1 test /Users/mike/oss/mapbox-gl-draw
> npm run lint && npm run tape


> @mapbox/mapbox-gl-draw@1.0.1 lint /Users/mike/oss/mapbox-gl-draw
> eslint --no-eslintrc -c .eslintrc index.js src test


> @mapbox/mapbox-gl-draw@1.0.1 tape /Users/mike/oss/mapbox-gl-draw
> tape -r ./test/mock-browser.js -r babel-register test/*.test.js

TAP version 13
# Draw.getFeatureIdsAt
```

I tried a fresh `npm install` with the lower npm version but that fails on the lack of `gulp-sourcemaps`.

mcwhittemore:
hmmm... I'm not really sure why this would happen.

gpbmike:
I got it working by looking at what Travis is doing. Turns out I needed to use node v6 and use yarn to install dependencies and then run tests.

Now on to the tests. I'm admittedly weak on tests. How would I go about spying on `setup.connect` when it's buried in a function response?

mcwhittemore:
I spent some time vetting this today. No obvious problems. We do need to refactor tests to make it easier to test this kind of change.

PR #687, opened 2017-09-12
Remove unnecessary setSelected call

### LINKED ISSUE #680: immediately double clicking with polygon tool breaks the map
**mapbox-gl-js version**: v0.39.1
**mapbox-gl-draw version**: v1.0.0

### Steps to Trigger Behavior

 1. enable polygon tool
 2. immediately double click the map

### Expected Behavior

No polygon should be drawn and polygon tool should continue to work.

### Actual Behavior

No polygon is drawn, error is thrown, and polygon tool can no longer be used.

Stack trace:
```
Uncaught TypeError: Cannot read property 'toGeoJSON' of undefined
    at src\render.js:57
    at Array.map (<anonymous>)
    at module.exports.render (src\render.js:57)
    at module.exports.wrapperFn [as render] (src\lib\throttle.js:21)
    at later (src\lib\throttle.js:8)
```

#340 was a similar issue

AliceR:
Fixes #680. 
The tests are still passing, but I am not sure about unwanted side-effects. Any feedback is appreciated :)

### DIFF
--- a/src/modes/mode_interface_accessors.js
+++ b/src/modes/mode_interface_accessors.js
@@ -16,8 +16,8 @@ const ModeInterface = module.exports = function(ctx) {
  * @name this.setSelected
  * @param {DrawFeature[]} - whats selected as a [DrawFeature](https://github.com/mapbox/mapbox-gl-draw/blob/master/src/feature_types/feature.js)
  */
-ModeInterface.prototype.setSelected = function(features, options = {}) {
-  return this._ctx.store.setSelected(features, options);
+ModeInterface.prototype.setSelected = function(features) {
+  return this._ctx.store.setSelected(features);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


mcwhittemore:
@AliceR - since the tests we're passing before, can you add a test that would have failed before you made this change?

AliceR:
I would need help for that. The test `'start draw_polygon mode then doubleclick'`(https://github.com/mapbox/mapbox-gl-draw/blob/c14acde0af91c236e4d67ac9ddb69df11c063fc2/test/draw_polygon.test.js#L599) should be failing (?), but it passes anyways. I couldn't find out how to test it.

mcwhittemore:
It looks like we're not testing for no event to be fired here. If we add a test for that than before this should be failing and after it would be good.

Its good to note that this is testing against an mock of mapbox-gl-js, so it won't act 100% like the real deal.

mcwhittemore:
Thanks @AliceR. I just quickly tested this locally and it looks good. I'll get a patch out ASAP.

mcwhittemore:
This has been published to npm in `v1.0.2`. Thanks again @AliceR.

PR #700, opened 2017-10-24
Fix some warnings

mike-marcacci:
I don't believe these have any impact on the code that gets installed/run, but appeases warnings that clutter my build logs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,16 +67,16 @@
     "unassertify": "^2.0.3"
   },
   "peerDependencies": {
-    "mapbox-gl": ">=0.27.0 <=0.40.0"
+    "mapbox-gl": ">=0.27.0 <=0.41.0"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


mcwhittemore:
Thanks for this. Can you update the references to the modules in the code base so the tests pass and then we'll be good to go.

mike-marcacci:
Ah yes – sorry about that. I also realized `yarn.lock` was going to need to reflect the changes. Simply running `yarn install` made a few more diffs, though, so it may have already been out of date (not sure if you guys mostly use yarn or npm at Mapbox).

mcwhittemore:
this is now part of [v1.0.3](https://github.com/mapbox/mapbox-gl-draw/releases/tag/v1.0.3) and has been published to npm.

mike-marcacci:
Awesome, thanks!

PR #719, opened 2017-12-06
use circle-stroke-* paint properties in default theme

andrewharvey:
This PR replaces the old style of doing circle strokes in GL JS with the newer `circle-stroke-*` paint properties.

I've bumped GL JS to 0.42 because although `circle-stroke-*` landed in 0.29.0, a the bug fix "Account for circle-stroke-width in queryRenderedFeatures https://github.com/mapbox/mapbox-gl-js/pull/5514" is important to have.

This does have the unfortunate side affect of not working with older GL JS versions, but anyone who can't upgrade their GL JS can keep using the current version of Draw.

Also nullifies #711.


andrewharvey:
It looks like theme.js still has styles for the static mode. Should these be moved out into https://github.com/mapbox/mapbox-gl-draw-static-mode ?

mcwhittemore:
@andrewharvey - it looks like I didn't move the static mode stuff to its own repo https://github.com/mapbox/mapbox-gl-draw-static-mode so we'd need to think about how to do that well. Might be worth ticketing out rather than doing in this PR.

mcwhittemore:
@andrewharvey - what gain do we get out of using this new property? Is it faster, less bug prone, more future proof? I ask because requiring mapbox-gl-js>=0.42 is a break change and I'd prefer to not make breaking changes unless there is a clear benefit to doing so.

andrewharvey:
> what gain do we get out of using this new property?

Simplicity. I wanted to make the circles larger for an application and find it easier to do when it's using circle-stroke.

> Might be worth ticketing out rather than doing in this PR.

Agreed, it should be in a different PR.

mcwhittemore:
@andrewharvey - thanks for the PR, but I don't think this is a good reason to make a breaking change. I'm happy to add a section to the docs linking to alternative stylesheets if you wanted to publish this version as a module in its own right.

andrewharvey:
@mcwhittemore no problem. I don't think it's worth a new module for such a minor change, so I'll just leave this branch around so if Draw bumps the required GL JS version, this PR can be reopened.

mcwhittemore:
Sounds good. Thanks @andrewharvey

PR #741, opened 2018-02-03
Add example of adding draw mode to draw

benrudolph:
This wasn't totally clear to me when I went about building a new draw mode. I realize that `defaultMode` and `modes` are documented in a different file, but it still tripped me up because it didn't quite click as to how I'd hook the mode in. Just  a suggestion, feel free to close or modify.

### DIFF
--- a/docs/MODES.md
+++ b/docs/MODES.md
@@ -49,12 +49,13 @@ LotsOfPointsMode.toDisplayFeatures = function(state, geojson, display) {
   display(geojson);
 };
 
-// Add the new draw mode to the mapbox Draw object
+// Add the new draw mode to the MapboxDraw object
 var draw = new MapboxDraw({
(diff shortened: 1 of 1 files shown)


docs/MODES.md
```diff
@@ -48,6 +48,14 @@ LotsOfPointsMode.onKeyUp = function(state, e) {
 LotsOfPointsMode.toDisplayFeatures = function(state, geojson, display) {
   display(geojson);
 };
+
+// Add the new draw mode to the mapbox Draw object
```
mcwhittemore: This would setup draw without any other modes and `lots_of_points` depends on `simple_select` which depends on `direct_select` so we should include those as well.

mcwhittemore:
Great idea. I left one comment about improve the addition.

Issue #745, opened 2018-02-19
Annotations doesn't show on map after removing and adding a new draw instance to map.


jithintomz:
mapbox-gl-js version : 0.44.1
mapbox-gl-draw version :1.0.4

### Steps to Trigger Behavior

 1.Remove existing draw instance.
 2.Add a new draw instance to the map.
 3.Add a  feature to the draw.

The added feature won't show until user interacts with draw.

https://codepen.io/jithintom/pen/WMMEPj

mcwhittemore:
Closed by #749

Issue #724, opened 2017-12-13
adding source prevents geometry added to draw from rendering


demiurg:
https://codepen.io/demiurge/pen/baNKLP?editors=1011

**mapbox-gl-js version**: 0.42.2
**mapbox-gl-draw version**: 1.0.4

### Steps to Trigger Behavior

 1. add a source
 2. add geometry to draw
 3. observe no geometry being rendered
 4. click draw square, observe geometry rendered

### Expected Behavior

should render

### Actual Behavior

does not render. However, if no source is added prior, OR moved below the draw code, things work as expected. This looks like a bug in order of something.

herrethan:
I'm running into this very issue too!
In case this helps enlighten the situation, my hacky workaround for the time being is:
```
var onSourceReady = function(e){
  if(e.sourceId == 'mapbox-gl-draw-cold' && e.isSourceLoaded){
    map.off('sourcedata', onSourceReady);
    // draw.add() will properly render polygons now
  }
}
map.on('sourcedata', onSourceReady);
```

herrethan:
looks like a race condition with sources or layer drawing of some kind, as I found wrapping the `draw.add()` in a loop checking for `map.loaded()` works

mcwhittemore:
@demiurg - can you confirm if this is fixed by 1.0.7?

demiurg:
Confirmed! Thanks a lot :)

PR #749, opened 2018-03-05
Race condition loading shapes

nigelsim:
As outlined in my support request, we are unable to make Mapbox Draw correctly render features added before the control has fully loaded, but there are no events or otherwise to properly indicate that the control is ready.

This PR forces the `ctx.store` to be dirty when doing the initial rendering following adding its layers.

The issue is that the `render` `cleanup()` function is being called even if the map isn't ready yet, and this function sets `store.isDirty=false`, meaning that any unrendered features are ignored.

I've also changed the order of operations for the `draw.render` event for a similar reason, so that changes made to the draw instance in response to this event aren't immediately forgotten.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -79,8 +79,8 @@ module.exports = function render() {
     });
   }
 
-  store.ctx.map.fire(Constants.events.RENDER, {});
   cleanup();
+  store.ctx.map.fire(Constants.events.RENDER, {});
 
   function cleanup() {
     store.isDirty = false;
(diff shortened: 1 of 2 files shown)


nigelsim:
This may resolve #745 and #724

mcwhittemore:
@nigelsim - sorry for the delayed follow up. I got a chance to sit down with this today and all looks well. I'm working on rolling a new release and show have this out shortly.

PR #759, opened 2018-03-29
Add Rotate Mode to mode docs

tempranova:
Adding a little mode I made for rotation of features.


mcwhittemore:
This is a really cool idea! Thanks for thinking to link it!

PR #763, opened 2018-04-05
Remove right margin added by button element

jpmoyn:
The `button` element adds `margin-right: 8px;` as a style to the draw control buttons. The CSS i added will remove that extra margin. It throws off the look of the buttons. Here are screenshots:

before (you can see the margin):
screen shot 2018-04-05 at 10 37 13 am

after:
screen shot 2018-04-05 at 10 38 26 am


jpmoyn:
Ah I didn't see the demo page. My mistake.

It is happening for me, but I am using it for a react app within the context of [material-ui](https://github.com/mui-org/material-ui) so that must be where the `button` style is coming from.

I will close this! Sorry for the misunderstanding.

mcwhittemore:
Thanks!

Issue #766, opened 2018-04-21
Eliminate dependence on Map#fire – breaking change in mapbox-gl-js


mollymerp:
mapbox-gl-js recently changed the function signature for `fire`, so our most recent beta release (https://github.com/mapbox/mapbox-gl-js/pull/6549) breaks compatibility with this library. see change log here https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.45.0-beta.1

mapbox-gl-draw shouldn't be relying on `Map#fire` because it is not a public method, but if that is impossible, another option would be to call `fire` with an object that has a `type` attribute `{type: 'error', message: 'somethingbad', //... rest of the payload properties }`. We may also consider making the `Event` class public so third party libraries like this can use that constructor.

samanpwbb:
Fixed by https://github.com/mapbox/mapbox-gl-draw/commit/7eba80cae2b2b816b06119c09678f055881ce609

tristen:
Mapbox GL Draw still draws heavy dependence on `map.fire` so I don't really see this as resolved.

jfirebaugh:
Yeah, to be clear, `map.fire` is private as of https://github.com/mapbox/mapbox-gl-js/commit/2b645718743c4a5f229be9ab90df2a77328fbcce. A resolution to this issue would be for gl-draw to implement its own eventing mechanism and fire events on the `MapboxDraw`.

I'd rather revert the changes in gl-js than pursue something like 7eba80cae2b2b816b06119c09678f055881ce609.

samanpwbb:
😅 definitely a premature close. We needed to get this fix in quickly for Studio tomorrow. The correct next action is to refactor gl-draw.

mcwhittemore:
To be clear here, private or not, I screwed up here and exposed a dependencies functionality as core to Draw. This is made worse since custom modes are required to use `this.map.fire` to trigger draw events which means that to fix this we need to make a breaking change to Draw and rework how modes report events back.

All of that said, we should make this change, I just don't have time to really work on it right now.

PR #771, opened 2018-05-01
Format map.fire call to match breaking change

tristen:
Taking the easy way out. A complete solution would re-write the draw tools API with its own `draw.on('create')` using something like EventEmitter. 

ref #766


mcwhittemore:
@tristen - I'd like to have this to not be a breaking change. Is it possible for us to know which version of `fire` is in mapbox-gl?

tristen:
> @tristen - I'd like to have this to not be a breaking change. Is it possible for us to know which version of `fire` is in mapbox-gl? 

This change still uses `fire`. It's just that its a private method now and argument signature has changed. The API should remain the same. I think this is close as we can get without dire breaking changes (although it does result in `peerDependencies` being pinned to `mapbox-gl-js@0.45.0`)

mcwhittemore:
This PR currently requires all users of Draw to update to `mapbox-gl@0.45.0-beta.1` which is the breaking change I'm noting.

tristen:
OK. I guess what we could do look at `mapboxgl.version` and format the argument differently.

mcwhittemore:
> we could do look at mapboxgl.version and format the argument differently.

Cool! This will probably set us up to do the EventEmitter option as well.

mollymerp:
thanks for taking this on @tristen – let me know if you need anything!

tristen:
I think this _still_ results in a breaking change. version would need to be passed as an option

```js
new MapboxDraw({
  version: mapboxgl.version
});
```

Which we can helpfully throw as an error if the user hasn't passed one?

tristen:
🚶 https://github.com/mapbox/mapbox-gl-draw/pull/772

PR #772, opened 2018-05-01
Patch map.fire to format arguments passed to it.

tristen:
Here's another way to do #771 that results in a much smaller PR

Ref #766


mollymerp:
🐒💗

Issue #760, opened 2018-03-31
High cpu usage during mousemoves - unnecessary render calls?  


ambientlight:
**mapbox-gl-js version**: 0.44.1
**mapbox-gl-draw version**: 1.0.3

I have noticed generally high CPU usage on mousemove in `simple_select` mode which comes from `DrawContext.store.render()` from [event delegate](https://github.com/mapbox/mapbox-gl-draw/blob/09dc7c590ed915f6f772d14e1ba222fdbf1f88bc/src/lib/mode_handler.js#L34-L50). 

Why does render needs to be performed on mousemove in `simple_select`. Isn't its purpose is just to [query features and update the cursor to indicate selection](https://github.com/mapbox/mapbox-gl-draw/blob/4be7d99f377d6063dc949e9ea75cd84538a9c1b8/src/events.js#L47)? Or am I missing something?

mcwhittemore:
@ambientlight - Yea, if there is a good way to skip the render here we should do that. Looking I don't see a clear way to simple way to achieve this without adding some sort of internal "skipRender" function. If you have an idea, please feel free describe here or show it in a PR. Perf improvements are always welcome :).

mcwhittemore:
@ambientlight - I see you did some work on [this in a fork](https://github.com/ambientlight/mapbox-gl-draw/commit/c1647034d7d29fe58ade6b76898e5ea9fd8a7654#diff-2650e3200d76013773ddf2af657f5a3dR82). While that commit would fix the problem for `simple_select` it wouldn't fix it for other modes such as `direct_select` or any custom none drawing mode.

ambientlight:
@mcwhittemore: for me most critical was to just get rid of unnecessary re-renders in `simple_select`. Since it accounted for vast majority usage time in my particular case I went with this hotfix of just not letting delegate calls to go through. I haven't used it for other modes as I didn't have time to get into details of how mapbox-gl-draw works internally, so I have created this issue to make sure I am not doing anything wrong with such approach. I was a bit confused since I didn't understand why render needed to be called for 'simple-select' actually. 

Maybe adding `requiresRender` flag and then having `render()` logic adjusted for it might be a reasonable way to approach this. Also I found that each `geoJSONSource.setData()` triggers renders so maybe memoizing query results may be another improvement that could be made to avoid unnecessary rerenders when let's say we are highlighting objects for the sake of selection.

hyperknot:
[...]

PR #783, opened 2018-05-11
Don't run render on unused event handlers

mcwhittemore:
Updates the mode to object code which converts an object following the mode interface into Draw's internal concept mode concept. The update makes sure that the selector for all events inherited from the interface is set to false so that the empty function code of the interface won't run.

This will reduce the amount of extra render calls as events that aren't directly implemented by a mode won't force renders anymore, but it does not solve #760 as any event handler implemented by a mode will still force a render even if that event handler does nothing to the source data.

This change means that the static custom mode will only force a render on setup.

@hyperknot if you can review that this solves your CPU problem in the static mode context that would be great.

### DIFF
--- a/src/modes/object_to_mode.js
+++ b/src/modes/object_to_mode.js
@@ -1,5 +1,22 @@
 const ModeInterface = require('./mode_interface');
 
+const eventMapper = {
+  drag: 'onDrag',
+  click: 'onClick',
+  mousemove: 'onMouseMove',
+  mousedown: 'onMouseDown',
+  mouseup: 'onMouseUp',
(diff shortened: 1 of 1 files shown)


hyperknot:
CPU usage is still very high for `simple_select`. StaticMode and `draw_point` is now fixed though.

Actually, you don't have to look at CPU usage, just console.log on `draw.render`, and you'll see that `simple_select` is firing renders all over the place.

mcwhittemore:
Thanks for the confirmation @hyperknot. This PR was only trying to fix it when a mode didn't even implement the event handler.

PR #798, opened 2018-06-17
Add link to radius mode

benrudolph:
Found this mode by @chriswhong to be very useful, so thought I'd link it in the repo. Can see it in action here: https://popfactfinder.planning.nyc.gov/


benrudolph:
Ah totally missed that issue, thanks. Some good points are brought up in that thread. FWIW I also use symbol layers in my mapbox-gl-draw modes as well as use them for selecting data on a map

Issue #725, opened 2017-12-13
Request: please provide ES5 compiled code


mgvarley:
**mapbox-gl-js version**: 0.42.2
**mapbox-gl-draw version**: 1.0.4

### Steps to Trigger Behavior

 1. Create a new react application using CreateReactApp https://github.com/facebookincubator/create-react-app
 2. npm install mapbox-gl and @mapbox/mapbox-gl-draw
 3. Create bare bones map using sample code for mapbox-gl-draw
 4. Attempt to build.  mapbox-gl only works fine and compiles, add mapbox-gl-draw to the mix and it breaks the build:

```
Creating an optimized production build...
Failed to compile.
 
Failed to minify the code from this file:

        ./node_modules/@mapbox/mapbox-gl-draw/src/modes/index.js:10

Read more here: http://bit.ly/2tRViJ9
```

Following the link suggests the issue is due to code not published as ES5 on npm, presumably mapbox-gl code is published in this way, would be great if mapbox-gl-draw would follow the same approach:

> npm run build fails to minify
> 
> Some third-party packages don't compile their code to ES5 before publishing to npm. This often causes problems in the ecosystem because neither browsers (except for most modern versions) nor some tools currently support all ES6 features. We recommend to publish code on npm as ES5 at least for a few more years.
> 
> To resolve this:
> Open an issue on the dependency's issue tracker and ask that the package be published pre-compiled.
> 
> 

### Expected Behavior

A clean build when including @mapbox/mapbox-gl-draw in a project that uses create-react-app and/or UglifyJS

### Actual Behavior

Including @mapbox/mapbox-gl-draw breaks the build.  Just including mapbox-gl works fine.

jingsam:
try import @mapbox/mapbox-gl-draw/dist/mapbox-gl-draw

mgvarley:
@ jingsam that worked perfectly, thank you!  Happy to submit a PR to add this into the README as a workaround but would be great if this worked as expected with `import @mapbox/mapbox-gl-draw`

rorysedgwick:
This is also causing a problem for me which is less simple to resolve, as we are using `react-mapbox-gl-draw`, which is requiring `@mapbox/mapbox-gl-draw` itself which we can't easily overwrite.

I realise this may be an issue to filed opened against the `react-mapbox-gl-draw` repo but do you have any suggestions for workarounds in the meantime?

mcwhittemore:
[...]

PR #806, opened 2018-08-07
include es5 versions of individual source files in build

amagee:
This pull request is basically a working proof of concept of this comment: https://github.com/mapbox/mapbox-gl-draw/issues/725#issuecomment-410929464

I haven't thought through all the issues but hopefully this can get a discussion going.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "lint": "eslint --no-eslintrc -c .eslintrc index.js src test",
     "tape": "tape -r ./test/mock-browser.js -r babel-register test/*.test.js",
     "coverage": "NODE_ENV=test nyc --reporter html npm run tape && opener coverage/index.html",
-    "build": "NODE_ENV=production browserify index.js --standalone MapboxDraw > dist/mapbox-gl-draw.js",
+    "build": "./scripts/build.sh",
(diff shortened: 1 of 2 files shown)


mcwhittemore:
@amagee - Thanks for pushing forward for this via a PR. My main concern for a PR that changes the build system is ensuring that the result is non-breaking for other users (we've failed at this in the past). How would your propose testing this to make sure it works in many situations?

amagee:
Yep good question.

1. People using the published npm packages
Assuming you run `npm run build` as part of your scripts to publish to npm, and the build succeeds, it should not be possible for this to break anything for anyone, since the main file generated by the build is unchanged; there are just new additional files. People who just import Mapbox Draw normally should be unaffected by the new files.

2. People running builds themselves
In its current form this could cause issues (eg. if run on a system that doesn't have `sed`, eg. Windows). This should be easily fixable though, if it's decided this is a sensible approach then I could replace the shell script with a Javascript file that does the same thing.

jmking:
Wondering if there is any more feedback on how we might proceed with this? (I work with @amagee)

kkaefer:
Superseded by https://github.com/mapbox/mapbox-gl-draw/pull/945

PR #815, opened 2018-09-06
Allow dry run of move_feature

fc:
If overriding the `dragMove` (simple_select)/`dragFeature` (direct_select) methods to create custom drag behavior, it is necessary to see what the new dragged coordinates are before actually dragging.  For example, I do not want the dragged feature(s) to be dragged outside of a set bounding area.

Rough example of usage:
```
import booleanContains from '@turf/boolean-contains';
import { polygon } from '@turf/helpers';

// ....

const delta = {
  lng: e.lngLat.lng - state.dragMoveLocation.lng,
  lat: e.lngLat.lat - state.dragMoveLocation.lat
};
const movedCoordinates = moveFeatures(this.getSelected(), delta, true);
const isInBounds = booleanContains(boundsPolygon, polygon(movedCoordinates[0]));
// then, perform drag for real
```

### DIFF
--- a/src/lib/move_features.js
+++ b/src/lib/move_features.js
@@ -1,10 +1,10 @@
 const constrainFeatureMovement = require('./constrain_feature_movement');
 const Constants = require('../constants');
 
-module.exports = function(features, delta) {
+module.exports = function(features, delta, isDryRun=false) {
   const constrainedDelta = constrainFeatureMovement(features.map(feature => feature.toGeoJSON()), delta);
 
-  features.forEach(feature => {
+  return features.map(feature => {
     const currentCoordinates = feature.getCoordinates();
 
     const moveCoordinate = (coord) => {
@@ -28,6 +28,9 @@ module.exports = function(features, delta) {
       nextCoordinates = currentCoordinates.map(moveMultiPolygon);
     }
 
-    feature.incomingCoords(nextCoordinates);
+    if (!isDryRun) {
+      feature.incomingCoords(nextCoordinates);
+    }
+    return nextCoordinates;
   });
 };


mcwhittemore:
`move_feature` is currently a public api of Draw if I remember correctly and it looks like this PR is suggestion that it should be. Is that a correct read? While I see the value of doing this, expanding the api surface of this project isn't something I think we're prepped to do. Are you looking to do this in a custom mode?

fc:
You are correct, I would like to expose it more but only because there is currently not a way to affect the existing drag behavior while in `simple_select`/`direct_select` if you want it to stop dragging.

Am I overlooking something?  Is there a better way to handle my particular use case?

If the custom mode API exposed what the next/new drag coordinates were to the `onDrag` then this would be a lot easier to handle.  Right now I get `dragMoveLocation` but that doesn't give me a chance to test whether the drop area can be cancelled or not.

I can't see how the above can be achieved easily since I'm not totally clear on where `onDrag` is called from? (maybe you can point me to the right direction?)

Alternatively what if we added a line here:
https://github.com/mapbox/mapbox-gl-draw/blob/5b2280817c901d93b05c937a7bdb2efb8afefe0c/src/modes/simple_select.js#L161

that maybe looks something like this:
```
state.dragNewCoords = moveFeatures(this.getSelected(), delta);

```
then in `dragMove()` would need to be modified too (both in direct/simple select).  Might be some work involved to get it right.

-----


This is how I'm using the custom mode at the moment.
```
const NewSimpleSelect = _.extend(MapboxDraw.modes.simple_select, {
  onDrag(state, e) {
    // stop drag if attempting to drag outside bounds using a modified `moveFeatures`
  }
};
```
I can achieve what I want by copying everything from `moveFeatures` and modifying it to fit my needs but this is not ideal.  I can't quite figure out how achieve what I want using a custom mode since I want to override something that is deeply baked into the default mode.  If I did create a fully custom mode it would be re-implementing most of what you have anyway.

mcwhittemore:
I'm not sure how well this will work inside of a custom mode, but I've enforced location limits before by listening to the `draw.update` event and resetting the feature to an old state if the new feature doesn't meet my constraints.

PR #817, opened 2018-09-06
Include userProperties in Options section of docs

fc:
While this is mentioned in the bottom of the API.md in the styling section, it was not included as part of the section that lists all of the available options.


mcwhittemore:
Thanks @fc!

Issue #617, opened 2017-03-31
map.addControl(draw) breaks touch events for popups


ebendennis:
Adding the draw controls to a map affects the functionality of touch events for popups. Click events still fire as expected, but touch events no longer fire.

mcwhittemore:
Eben,

Thanks for bringing this to our attention. Cn you please provide a code example so we can see exactly how it breaks.

Regards,
Matthew

ebendennis:
Matthew,

Been working on this app where I noticed the issue: [https://github.com/iconengineering/maps/tree/master/comments-test](https://github.com/iconengineering/maps/tree/master/comments-test)

mcwhittemore:
Eben,

Thanks for this. One thing I noticed was that you are loading Draw inside of `map.on('load' ...);` This might not affect this problem, but its pretty important.

Also, if possible, can you wire up something in jsbin. I need to review a bunch of issues and PRs in a single day, and being able to quickly test them out will help me respond to you. Sorry, I should have said this in my first comment.

Regards,
Matthew

ebendennis:
No problem, here you go - [https://jsbin.com/runaxuv/edit?js,output](https://jsbin.com/runaxuv/edit?js,output)

mcwhittemore:
Eben,

Thanks for the jsbin. I just play around with this a bit and I see what you mean. My guess is that Draw is stopping all touch events even if it takes no action on the event. We probably need to make sure `stopExtendedActions` doesn't happen if there are more dom nodes that could handle the event.

https://github.com/mapbox/mapbox-gl-draw/blob/e1df892d543977e02265c7a9f634fc97d076f83b/src/modes/simple_select.js#L117-L126

jwilliams-ecometrica:
Hi @mcwhittemore, 

I'm just wondering has there being any progress on resolving this issue?

mcwhittemore:
@jwilliams-ecometrica, nope. If you have time to help fix it, that would be great!

jwilliams-ecometrica:
Hi @mcwhittemore, Fair enough. I'm not sure quite how to go about fixing this (Other than simply removing the offending call to stopExtendedInteractions().

mcwhittemore:
`stopExtendedInteractions()` is being called in too many cases here. Ideally this would only run if the map was the direction of the tap, right now its running when its child elements are clicked as well.

mcwhittemore:
[...]

PR #830, opened 2018-10-26
Enable touch event propagation

Plantain:
Touch events are inexplicably prevent from propagating while mouse clicks propagate fine.
Fix for #617

Requires testing across some more devices.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -78,9 +78,6 @@ module.exports = function(ctx) {
   };
 
   events.touchstart = function(event) {
-    // Prevent emulated mouse events because we will fully handle the touch here.
-    // This does not stop the touch events from propogating to mapbox though.
-    event.originalEvent.preventDefault();
     if (!ctx.options.touchEnabled) {
       return;
     }


raysuelzer:
Please merge this.

sergei-zelinsky:
Are there any updates on this PR? Will it be merged or not?

Thanks!

rreusser:
This looks closely related to #869, with the exception that that PR moves this until later instead of preventing default altogether. I think this PR can be closed in favor of that one.

m-vo:
Is there anything preventing this PR from being merged + released?

Just for the record: This bug makes using mapbox virtually impossible with Chrome devtools opened in 'touch device mode' and I fear the same is true for all devices supporting both mouse + touch. Placing and moving a point for instance works, but no `draw.update` events will be fired on touch interaction. Bummer. :cry:

stepankuzmin:
We've switched to passive event listeners instead of `preventDefault` here https://github.com/mapbox/mapbox-gl-draw/pull/1146

PR #847, opened 2018-12-14
Pass all properties when creating midpoints and vertices, close #846

### LINKED ISSUE #846: Pass userProperties to midpoints and vertices
On creation of a midpoint or a vertex, the new feature is initialized with a reference to the parent feature but none of the properties are passed down. This means that midpoints and vertices can't be styled based on user-provided properties. I'd like to work up a PR where the user-provided properties are passed down to these features for potential styling. Thoughts?

Parent references:
- https://github.com/mapbox/mapbox-gl-draw/blob/5b2280817c901d93b05c937a7bdb2efb8afefe0c/src/lib/create_vertex.js#L19
- https://github.com/mapbox/mapbox-gl-draw/blob/5b2280817c901d93b05c937a7bdb2efb8afefe0c/src/lib/create_midpoint.js#L24

drewbo:
Updated the `createMidpoint` and `createVertex` to take a first parameter GeoJSON feature instead of an id. This allows passing properties down to the created points. This also updates all the calls to those functions (I think) and the necessary tests (plus two extra for adding `userProperties`)

### DIFF
--- a/src/lib/create_midpoint.js
+++ b/src/lib/create_midpoint.js
@@ -1,5 +1,15 @@
 const Constants = require('../constants');
 
+/**
+ * Returns GeoJSON for a Point representing the
+ * midpoint of another feature.
+ *
+ * @param {GeoJSON} parent
+ * @param {GeoJSON} startVertex
+ * @param {GeoJSON} endVertex
+ * @param {Object} map
+ * @return {GeoJSON} Point
+ */
 module.exports = function(parent, startVertex, endVertex, map) {
   const startCoord = startVertex.geometry.coordinates;
   const endCoord = endVertex.geometry.coordinates;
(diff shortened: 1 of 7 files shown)


awulkan:
I have also created an implementation for this, which works with the latest version of Mapbox Draw, and can be seen here:
https://github.com/awulkan/mapbox-gl-draw/tree/vertex-midpoint-properties

I will not submit a pull request though since it appears that @drewbo has already submitted one. I didn't know that this pull request existed until I was about to submit my own.

**Please consider** adding this functionality from either my or drewbo's code. It's a highly necessary feature for anyone that needs custom styling. In my situation I need to be able to hide vertices and midpoints so that our users don't think they can drag/reshape certain polygons.

Related issues: #846

Issue #823, opened 2018-09-28
Compatibility with Firefox  and IE 11


thu07:
Thank you so much for your great work. However, when I integrate @mapbox-gl-draw into my angular 6 project and run in Firefox and IE 11, it exists unbelievable errors.

In Firefox, the error is "SyntaxError: missing = in const declaration"; in IE 11, the error is “SCRIPT1028: Expected identifier, string or number”. I try to find the problem, however, I can't find the error in the places where the browser points out. 

Meanwhile, I have set up the compatibility configuration file in angular. Without importing @mapbox-gl-draw, Chrome, Firefox, and IE 11 works well. By the way, the Mapbox-gl version is 0.49.0 and Mapbox-gl-draw version is 1.0.9. The angular cli is the latest version as well.

Hope someone can help me to fix this problem. Appreciate it.

best

Tao

andrewharvey:
It sounds like you need to either use the prebuilt bundle or pass it through babel?

thu07:
> It sounds like you need to either use the prebuilt bundle or pass it through babel?

Thank you so much for your reply. However, I try to use webpack and babel to compile the code, it still does not work. 

The .babelrc file is like:
{
  "presets": [
    "@babel/preset-env",
    "@babel/preset-typescript"
  ]
}

The webpack.config.js file is like:

const path = require('path');

module.exports = {
  entry: { main: './src/index.ts' },
  output: {
    path: path.resolve(__dirname, 'dist'),
    filename: 'output.js'
  },
  resolve: {
    extensions: ['.js', '.ts']
  },
  module: {
    rules: [
      {
        test: /\.ts$/,
        exclude: /node_modules/,
        use: {
          loader: 'babel-loader'
        }
      },
      {
        test: /\.html$/,
        use: [
          {
            loader: "html-loader"
          }
        ]
      }
    ]
  },
  "node": {
    "fs": "empty"
  }
};

And the app.component.ts file is like:

import { Component, OnInit } from '@angular/core';
import * as mapboxgl from 'mapbox-gl';
import * as MapboxDraw from '@mapbox/mapbox-gl-draw';

@Component({
  selector: 'app-root',
  // template: `
  //   <div>
  //     <h1>{{title}}</h1>
  //     <div>The number: {{x}}</div>
  //   </div>
  // `,
  template: `
    <div id='map' class='map'>    
    </div>
  `,
})
export class AppComponent implements OnInit {
  title: string = 'Working component!';
[...]

PR #854, opened 2019-01-10
Fixed old version browser const issue

agiratech-manigandan:
I am getting syntax Error for old version browsers

Error Message:

SyntaxError: missing = in const declaration

I checked why this is coming . https://github.com/mapbox/mapbox-gl-draw/blob/master/src/feature_types/feature.js in this file const using the loop

if (this.ctx.options.userProperties) {
for (const name in this.properties) {
properties[user_${name}] = this.properties[name];
}
}

Could to change to const to let.

Reference : https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Errors/Missing_initializer_in_const

Fix for #823

### DIFF
--- a/src/feature_types/feature.js
+++ b/src/feature_types/feature.js
@@ -52,7 +52,7 @@ Feature.prototype.internal = function(mode) {
   };
 
   if (this.ctx.options.userProperties) {
-    for (const name in this.properties) {
+    for (let name in this.properties) {
       properties[`user_${name}`] = this.properties[name];
     }
   }


kkaefer:
This was fixed by switching to buble/rollup in https://github.com/mapbox/mapbox-gl-draw/commit/69dd2b6e50d27264b1d79b608c944499790a73b3

kkaefer:
Note that there's no release yet that includes this fix.

PR #870, opened 2019-02-22
refactor(render.js): prevent internal updates to map drawing sources

goldpbear:
…urceData method

This PR refactors map source updates from within `mapbox-gl-draw` such that they don't follow the `getSource().setData()` pattern but rather call a single, top-level method called `setSourceData` which is designed to be monkey-patched on the map instance at runtime.

The purpose of this is to reroute what would otherwise be internal state updates to React/Redux state so `react-map-gl` can remain stateless while using `mapbox-gl-draw`.

To be clear this is pretty hacky. We should replace this whole setup when `react-map-gl` introduces support for a drawing overlay.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -41,16 +41,12 @@ module.exports = function render() {
   }
 
   if (coldChanged) {
-    store.ctx.map.getSource(Constants.sources.COLD).setData({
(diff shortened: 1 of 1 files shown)


goldpbear:
Sorry, totally opened this PR in the wrong place...

Issue #698, opened 2017-10-20
Clicking draw button toggles active class, doesn't cancel drawing mode


andycochran:
The `.active` class on the draw buttons can't be used to indicate to the user that they are in drawing mode — e.g. by changing the color of the button — because clicking the draw button a second time does not cancel drawing mode. 

### Steps to Trigger Behavior

1. Go to [mapbox.com/mapbox-gl-js/example/mapbox-gl-draw](https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/)
2. Inspect the `.mapbox-gl-draw_polygon` button 
3. Click the polygon button (note in the inspector that the button gets an `.active` class) 
4. Click the polygon button again (note the `.active` class is removed) 
5. Draw a polygon on the map

### Expected Behavior
If clicking the draw button toggles its `.active` class, the draw mode should also be toggled. The active class should be used to indicate to the user that they are in drawing mode. 

### Actual Behavior
Clicking the draw button toggles its `.active` class but does not toggle drawing mode. So the active class can't be used to indicate drawing mode. 

**mapbox-gl-js version**: 0.40.1
**mapbox-gl-draw version**: 1.0.2

jesusr:
+1 This is the behaviour I am looking for.

huaruiwu:
Any reason why my PR isn't being looked at or accepted?

kkaefer:
PR got merged. Thanks for your contribution!

PR #872, opened 2019-02-25
draw buttons should cancel on toggle

huaruiwu:
Relates to this issue https://github.com/mapbox/mapbox-gl-draw/issues/698. The buttons should actually toggle the mode you're in since it toggles `active` css class. If this makes sense I can modify the tests

### DIFF
--- a/src/ui.js
+++ b/src/ui.js
@@ -69,6 +69,7 @@ module.exports = function(ctx) {
       const clickedButton = e.target;
       if (clickedButton === activeButton) {
         deactivateButtons();
+        options.onDeactivate();
         return;
       }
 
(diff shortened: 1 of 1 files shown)


huaruiwu:
Any reason why my PR isn't being looked at or accepted?

ansis:
Thanks for fixing this! Our maintenance of this library has not been where it should be. Thanks for the patience

huaruiwu:
No problems, thank you for maintaining the library!

krvajal:
Is this fix released?

crow7m:
Was this released ? pretty important now
Thank you for maintaining

kkaefer:
You can check out a beta release that includes this fix at https://github.com/mapbox/mapbox-gl-draw/releases/tag/v1.2.0-beta.1. It's also available via NPM.

Issue #842, opened 2018-11-25
Typescript type declarations?


maxburke:
Are there plans in the near future for supplying a Typescript type declarations file? 

(Presently using mapbox-gl-draw 1.0.9)

vanhumbeecka:
While awaiting the official typescript declaration files, I've written the basics manually, based on the readme from this repo. Use with care. This can still contain mistakes:
```
declare module '@mapbox/mapbox-gl-draw' {
  import {Feature, FeatureCollection} from 'geojson'
  import IMapboxDrawControls = MapboxDraw.IMapboxDrawControls

  class MapboxDraw {
    constructor(options?: {
      displayControlsDefault?: boolean,
      keybindings?: boolean,
      touchEnabled?: boolean,
      boxSelect?: boolean,
      clickBuffer?: number,
      touchBuffer?: number,
      controls?: IMapboxDrawControls,
      styles?: object[],
      modes?: object,
      defaultMode?: string,
      userProperties?: boolean
    });

    public add(geojson: object): string[]

    public get(featureId: string): Feature | undefined

    public getFeatureIdsAt(point: { x: number, y: number }): string[]

    public getSelectedIds(): string[]

    public getSelected(): FeatureCollection

    public getSelectedPoints(): FeatureCollection

    public getAll(): FeatureCollection

    public delete(ids: string | string[]): this

    public deleteAll(): this

    public set(featureCollection: FeatureCollection): string[]

    public trash(): this

    public combineFeatures(): this

    public uncombineFeatures(): this

    public getMode(): string

    public changeMode(mode: string, options?: object): this

    public setFeatureProperty(featureId: string, property: string, value: any): this

  }

  export = MapboxDraw
}

declare namespace MapboxDraw {
  export interface IMapboxDrawControls {
    point?: boolean,
    line_string?: boolean,
    polygon?: boolean
    trash?: boolean,
    combine_features?: boolean,
    uncombine_features?: boolean
  }
}
```

amaurymartiny:
@vanhumbeecka To get more visibility I think you can create a PR to create an `index.d.ts` file with these typings, it's already a start.

vanhumbeecka:
@amaurymartiny done!

Tyler-V:
Why hasn't this been merged in yet?

maxburke:
@amaurymartiny is it possible to merge the change from @vanhumbeecka ?

maxburke:
[...]

PR #877, opened 2019-03-20
Typescript definition file added

vanhumbeecka:
Related Github issue: #842

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "author": "mapbox",
   "license": "ISC",
   "main": "index.js",
+  "types": "index.d.ts",
   "style": "dist/mapbox-gl-draw.css",
   "browserify": {
     "transform": [
--- a/package.json
+++ b/package.json
@@ -76,12 +76,12 @@
   },
   "dependencies": {
     "@mapbox/geojson-area": "^0.2.1",
+    "@mapbox/geojson-extent": "^0.3.2",
     "@mapbox/geojson-normalize": "0.0.1",
     "@mapbox/geojsonhint": "^2.0.0",
-    "@mapbox/geojson-extent": "^0.3.2",
+    "@mapbox/point-geometry": "0.1.0",
     "hat": "0.0.3",
     "lodash.isequal": "^4.2.0",
-    "@mapbox/point-geometry": "0.1.0",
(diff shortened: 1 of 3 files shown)


amaurymartiny:
I believe you should add a `types` field in package.json pointing to index.d.ts: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

package.json
```diff
@@ -59,6 +59,7 @@
     "eslint": "^4.2.0",
     "eslint-config-mourner": "^2.0.1",
     "express": "^4.13.4",
+    "geojson": "^0.5.0",
```
amaurymartiny: This should be added via `yarn add geojson`, right now the yarn.lock is not updated

vanhumbeecka: Got it

ansis:
Hey! Thanks for your interest and contributing these. Since we can't commit to maintaining these definitions I think it would be best for us not to merge it. Previously we have had users using mapbox-gl commit typescript definitions for that to DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/11386). Would that approach work for you?

I'd be happy to merge a change to the readme that mentions that typescript definitions are available through DefinitelyTyped.

kkaefer:
Yep, DefinitelyTyped is the where these should go. Thanks for your contribution!

JamesLMilner:
@vanhumbeecka did you ever get a chance to raise this on DefinitelyTyped? I couldn't see anything in the history: https://github.com/DefinitelyTyped/DefinitelyTyped/pulls?utf8=%E2%9C%93&q=is%3Apr+mapboxgl

vanhumbeecka:
@JamesLMilner no i couldn't. So far I still just use these hardcoded in a separate `types` folder and point my `tsconfig.json` file to it for extra type source. Works well enough.

JamesLMilner:
@vanhumbeecka okay, thanks for getting back to me. Any chance you could share your up to date types here or in a Gist perhaps? I could look into raising it longer term

chriszrc:
I'm not seeing this in DefinitelyTyped:

https://github.com/DefinitelyTyped/DefinitelyTyped/search?q=mapboxdraw

Is this up yet? Is there some reason it can't be added?

PR #884, opened 2019-04-17
feature: add custom controls with `constrols` prop

Akiyamka:
When we create custom modes we can't add custom control for that. That request add ability add custom controls.


Akiyamka:
Please up node version in CI to  ^7.0.0
Node 10.15.3 - LTS at this moment .

Issue #889, opened 2019-05-22
Listed as not compatiable with mapbox-gl 1.0.0 that's required for new pricing model


daveisfera:
**mapbox-gl-js version**: 1.0.0
**mapbox-gl-draw version**: 1.1.1

### Steps to Trigger Behavior

 1. Update to `mapbox-gl-js` for new pricing
 2. Run `yarn install`
 3. Observe warning

### Expected Behavior
Mapbox's products are compatible with the requirement for the new pricing model

### Actual Behavior
An unexpected error that doesn't make sense:
```
warning "web > @mapbox/mapbox-gl-draw@1.1.1" has incorrect peer dependency "mapbox-gl@>=0.27.0 <=0.51.0".
```

ansis:
I've published @mapbox/mapbox-gl-draw@1.1.2 with an updated peer dependency

PR #890, opened 2019-05-22
update peer dependency

ansis:
fix https://github.com/mapbox/mapbox-gl-draw/issues/889


mcwhittemore:
Thanks for doing this @ansis.

PR #901, opened 2019-06-29
Fix selection event in direct_select w/ coordPath

Zirak:
The `draw.selectionchange` event fired when moving into direct_select mode with
a `coordPath` will now correctly include the point. Before, the event was
fired with the correct selected feature, but lacking in the point.

This bug was caused because of a misordered selection call: First the
coordinates were selected, and only then the feature. During the feature
selection, the store's `refreshSelectedCoordinates` function was called
*before* the feature was actually selected, meaning the previously
coordinates were effectively removed.

### DIFF
--- a/src/modes/direct_select.js
+++ b/src/modes/direct_select.js
@@ -132,8 +132,8 @@ DirectSelect.onSetup = function(opts) {
     selectedCoordPaths: opts.coordPath ? [opts.coordPath] : []
   };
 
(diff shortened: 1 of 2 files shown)


DannyDelott:
Nice! This fixes an issue when changing into `direct_select` mode w/ a coordPath that the user then wants to delete. 

@kkaefer can this be merged?

usb248:
+1

Issue #886, opened 2019-04-19
Is there any method or parameter that makes the drawn graphic impossible to drag?


The-End-Hero:

awulkan:
I'm also interested in knowing this. Do we have to create our own mode for it?

ryanhamley:
You can use the `Static` custom mode via [this plugin](https://github.com/mapbox/mapbox-gl-draw-static-mode).

awulkan:
@ryanhamley I though static mode disabled all interaction, not just the ability to drag things.

awulkan:
Any news regarding this? Is it still impossible to just disable the ability to drag a feature?

Issue #753, opened 2018-03-12
Disabling dragging/moving of drawn features


nigelsim:
When drawing a feature it is possible to drag and move the entire feature. This is confusing to users as they expect to be dragging the map, so we want to be able to disable this functionality.

**mapbox-gl-js version 0.44.1**:
**mapbox-gl-draw version 1.0.4**:

### Steps to Trigger Behavior

 1. Draw a polygon
 2. Click on the polygon
 3. When hovering over the polygon the cursor becomes a 'move' icon
 4. Clicking and dragging moves the polygon

### Expected Behavior

No move cursor and polygon not moved.

### Actual Behavior

Cursor changes and polygon is movable.

mcwhittemore:
This is expected behavior of how mapbox-gl-draw works. For custom interactions please see [Custom Modes](https://github.com/mapbox/mapbox-gl-draw/blob/master/docs/MODES.md#writing-custom-modes).

clementmas:
Can anyone provide an example of how to do this with a Custom Mode? Thanks :)

scooterGrun:
Have a look at https://github.com/mapbox/mapbox-gl-draw-static-mode

stevage:
See also solution at https://github.com/mapbox/mapbox-gl-draw/issues/667#issuecomment-1707799304

vcardins:
Although It's an old thread but it's exactly the problem I'm facing. How could I still drag the map without dragging/moving the feature?
I came across a solution that uses `map.panTo(e.lngLat);` however the transition is awkward. Ideas?

![2023-12-11 17 55 57](https://github.com/mapbox/mapbox-gl-draw/assets/2383329/a426c06a-e5b4-4a90-a889-ff145490441d)

PR #909, opened 2019-08-09
add canDrag option

vcebotari:
add `canDrag` option

* https://github.com/mapbox/mapbox-gl-draw/issues/886
* https://github.com/mapbox/mapbox-gl-draw/issues/753


slobglob:
@vcebotari Why wasn't this merged eventually?

PR #922, opened 2019-10-08
allow to dynamically update the drawing controls

### LINKED ISSUE #824: Disable/Remove Controls
Hi, 

I think that allowed the possibility to disable or remove controls could be interesting.

For example you want the user to create a polygon or a point but only one and still let him use update function or delete ...
Actually as far as i know you can't block drawing action that mean he can create as many polygon he wants and i can't find a way to update controls.

krvajal:
Exposes a new method called `setControls` to dynamically toggle the
drawing controls

Example usage
```javascript
 const drawControl = new MapboxDraw({
            displayControlsDefault: false,
            controls: {
                polygon: true,
                trash: true,
            },
        });
// add the control to the map
map.addControl(drawControl, 'bottom-right');

// now show only the trash 
drawControl.setControls({trash: true})

```
This is my first PR to this project and not sure if I missing some necessary step. I would appreciate any help to get this merged as it is a very handy feature.

fixes #824

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -192,5 +192,11 @@ module.exports = function(ctx, api) {
     return api;
   };
 
+  api.setControls = function(controls) {
+    ctx.options.controls = controls;
+    ctx.ui.removeButtons();
+    ctx.ui.addButtons();
+  };
+
   return api;
 };
--- a/src/setup.js
+++ b/src/setup.js
@@ -6,7 +6,7 @@ const xtend = require('xtend');
 
 module.exports = function(ctx) {
 
-  let controlContainer = null;
+
   let mapLoadedInterval = null;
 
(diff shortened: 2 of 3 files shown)


JClackett:
Please merge this 👍

lupas:
I'd also vote for this to be merged, would be super useful :)

I need to restrict users to only create 1 element and not multiple, and this would allow me to do that.

wtravO (approved):

isaac-souza:
I posted a possible workaround here https://github.com/mapbox/mapbox-gl-draw/issues/824#issuecomment-1727854277 if anyone is still looking at this

crow7m:
Hey, any update on this will be really really useful , thank you

Issue #897, opened 2019-06-17
Deleting Vertices - Wrong Points Removed on Trash


mikeomeara1:
**mapbox-gl-js version**: 1.0
**mapbox-gl-draw version**: 1.0.9

### Steps to Trigger Behavior

 1. Go to:  https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/
 2. Draw a polygon with more than ~20 points
 3. Select more than 10 vertices starting at the 10th point or later
 4. Click the "Trash" button
 4. At least one (in cases with polygons that have many vertices - many) of the selected points still remain while some un-selected points are removed.

### Expected Behavior
All **selected** points are removed when in `direct_select` mode and the `trash` event is executed.

### Actual Behavior
Some selected points will still remain. Some un-selected points are removed.

### Example 1 - Large number of points
![image](https://user-images.githubusercontent.com/8380449/59642403-b3e74980-9119-11e9-9e7e-ffc9b4d617b2.png)
Yields the following after `trash`:
![image](https://user-images.githubusercontent.com/8380449/59643278-eba3c080-911c-11e9-8135-7cf63b613d1e.png)

### Example 2 - Small number of points
![image](https://user-images.githubusercontent.com/8380449/59643105-586a8b00-911c-11e9-8b23-707be19d5ee2.png)
After Trash:
![image](https://user-images.githubusercontent.com/8380449/59643243-c020d600-911c-11e9-8c75-17d45ac23bdd.png)

### Remediation
I'm not certain, but I believe this has something to do with the `coord_path` assigned to every vertex. When the `trash` event fires on `direct_select` it does:
```javascript
state.selectedCoordPaths.sort().reverse().forEach(id => state.feature.removeCoordinate(id));
```
I believe that the `sort()` function doesn't take into account that `coord_path` is a string and when the values get greater than `"0.10"` the sort function puts the vertices in the wrong order, thus removes the wrong ones.

mikeomeara1:
After troubleshooting, it appears if you change the following line in `direct_select.js` (line 168) from:

```javascript
state.selectedCoordPaths.sort().reverse().forEach(id => state.feature.removeCoordinate(id));
```

To:

```javascript
 state.selectedCoordPaths.map(p => {return {coord_path: p, sort_val: parseInt(p.split('.')[p.split(".").length -1], 10)}})
    .sort((a, b) => a.sort_val - b.sort_val)
    .map(p => p.coord_path)
[...]

PR #943, opened 2019-12-18
Trash correct vertices by changing sort to be numeric-aware

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-draw/issues/897 by correctly sorting coordinate paths prior to deletion.

We're using "coordinate paths" to store which vertices are selected. These paths are stringified indices into the coordinates array, and can be multiple levels deep separated by `.`. When deleting vertices, we sort them in reverse order to make sure that we invalidate the indices by deleting vertices that come before another vertices scheduled for deletion. This works if we have fewer than 10 vertices. However, the alphabetical sort means that we'd delete note `10` before node `9`. This PR fixes this by using a numeric sort.

### DIFF
--- a/src/modes/direct_select.js
+++ b/src/modes/direct_select.js
@@ -165,7 +165,11 @@ DirectSelect.toDisplayFeatures = function(state, geojson, push) {
 };
 
 DirectSelect.onTrash = function(state) {
-  state.selectedCoordPaths.sort().reverse().forEach(id => state.feature.removeCoordinate(id));
+  // Uses number-aware sorting to make sure '9' < '10'. Comparison is reversed because we want them
+  // in reverse order so that we can remove by index safely.
+  state.selectedCoordPaths
+    .sort((a, b) => b.localeCompare(a, 'en', { numeric: true }))
+    .forEach(id => state.feature.removeCoordinate(id));
   this.fireUpdate();
   state.selectedCoordPaths = [];
   this.clearSelectedCoordinates();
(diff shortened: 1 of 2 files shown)


src/modes/direct_select.js
```diff
@@ -165,7 +165,11 @@ DirectSelect.toDisplayFeatures = function(state, geojson, push) {
 };
 
 DirectSelect.onTrash = function(state) {
-  state.selectedCoordPaths.sort().reverse().forEach(id => state.feature.removeCoordinate(id));
+  // Uses number-aware sorting to make sure '9' < '10'. Comparison is reversed because we want them
+  // in reverse order so that we can remove by index safely.
+  state.selectedCoordPaths
+    .sort((a, b) => b.localeCompare(a, 'en', { numeric: true }))
```
mourner:
```suggestion
    .sort((a, b) => +b - a)
```

kkaefer: This doesn't work for sort keys that are deeply nested, e.g. `0.8.9` should be sorted before `0.8.10`

kkaefer: I tested that this works with localeCompare's numeric sort:

```js
console.warn(['11', '10', '9', '33', '8', '30', '3'].sort((a, b) => b.localeCompare(a, 'en', { numeric: true })));
console.warn(['4.11', '4.10', '40.8', '4.9', '4.33',].sort((a, b) => b.localeCompare(a, 'en', { numeric: true })));
console.warn(['4.10.34', '4.10.3', '4.10.29'].sort((a, b) => b.localeCompare(a, 'en', { numeric: true })));
```

mourner: Oh, I didn't realize those weren't just numbers, which wasn't clear from the comment. 👍

PR #944, opened 2019-12-18
Skip rerenders for mousemove handlers when in selection modes

kkaefer:
Reduces CPU usage and fixes https://github.com/mapbox/mapbox-gl-draw/issues/760

CPU usage is still high when moving the mouse, but the remaining CPU usage is in queryRendereredFeatures which is used to determine the shape of the mouse pointer.

### DIFF
--- a/src/lib/mode_handler.js
+++ b/src/lib/mode_handler.js
@@ -37,8 +37,10 @@ const ModeHandler = function(mode, DrawContext) {
     while (iHandle--) {
       const handle = handles[iHandle];
       if (handle.selector(event)) {
-        handle.fn.call(ctx, event);
-        DrawContext.store.render();
+        const skipRender = handle.fn.call(ctx, event);
+        if (!skipRender) {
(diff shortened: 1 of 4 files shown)


robhawkes:
This is great, would love to see it get merged so I don't need to run a custom build of draw. 👍

JamesLMilner:
Would also love to see this merged if possible :) I think it'd be a great improvement.

jayarjo:
Any idea when there will be a release with this?

usb248:
I tested this PR and it causes a bug.
These lignes of code : https://github.com/mapbox/mapbox-gl-draw/pull/944/files#diff-2650e3200d76013773ddf2af657f5a3dR40-R43.
On the first click (quick mousedown+mouseup) on a feature, the feature doesn't enter in selection owing to a "glitch" (re-render ?).

Please **fix this bug** or **not merge** this PR.
thanks

PR #945, opened 2019-12-20
first step of ES6 conversion + browserify -> rollup

kkaefer:
also made benchmarks functional again; some are still a bit buggy though.


kkaefer:
> There seem to be a ton of `export default { ... a singleton with a bunch of properties ... }` which could possibly be refactored so that each of the properties is a named export (where possible)

Yeah, I'm aware of that; it's a result of my semi-automated conversion. My plan was to leave this for a follow-up PR in the interest of making this PR easier to review and merge.

PR #952, opened 2020-01-02
Remove condition preventing vertex removal

jorgedl:
Removing a condition that checks if the trash button is wanted in the UI. The vertex removal on 'keydown' should not be related with the button rendering or not


jebenois:
Any news on that one?

PR #964, opened 2020-02-12
Pass userProperties to midpoints and vertices 

### LINKED ISSUE #846: Pass userProperties to midpoints and vertices

JamesLMilner:
This PR brings @drewbo PR for this issue (#847) up to speed with master. The aim here is to pass userProperties to midpoints and vertices, and in turn allow them to be styled. 

Resolves #846

### DIFF
--- a/src/lib/create_midpoint.js
+++ b/src/lib/create_midpoint.js
@@ -1,27 +1,40 @@
-import * as Constants from '../constants';
+import * as Constants from "../constants";
 
+/**
+ * Returns GeoJSON for a Point representing the
+ * midpoint of another feature.
+ *
+ * @param {GeoJSON} parent
+ * @param {GeoJSON} startVertex
+ * @param {GeoJSON} endVertex
+ * @param {Object} map
+ * @return {GeoJSON} Point
+ */
 export default function(parent, startVertex, endVertex, map) {
   const startCoord = startVertex.geometry.coordinates;
   const endCoord = endVertex.geometry.coordinates;
 
   // If a coordinate exceeds the projection, we can't calculate a midpoint,
   // so run away
-  if (startCoord[1] > Constants.LAT_RENDERED_MAX ||
+  if (
+    startCoord[1] > Constants.LAT_RENDERED_MAX ||
     startCoord[1] < Constants.LAT_RENDERED_MIN ||
     endCoord[1] > Constants.LAT_RENDERED_MAX ||
-    endCoord[1] < Constants.LAT_RENDERED_MIN) {
+    endCoord[1] < Constants.LAT_RENDERED_MIN
+  ) {
     return null;
   }
 
-  const ptA = map.project([ startCoord[0], startCoord[1] ]);
-  const ptB = map.project([ endCoord[0], endCoord[1] ]);
-  const mid = map.unproject([ (ptA.x + ptB.x) / 2, (ptA.y + ptB.y) / 2 ]);
+  const ptA = map.project([startCoord[0], startCoord[1]]);
+  const ptB = map.project([endCoord[0], endCoord[1]]);
+  const mid = map.unproject([(ptA.x + ptB.x) / 2, (ptA.y + ptB.y) / 2]);
 
-  return {
+  const midpoint = {
     type: Constants.geojsonTypes.FEATURE,
     properties: {
+      ...parent.properties,
       meta: Constants.meta.MIDPOINT,
-      parent,
+      parent: parent.properties && parent.properties.id,
       lng: mid.lng,
       lat: mid.lat,
       coord_path: endVertex.properties.coord_path
@@ -31,4 +44,6 @@ export default function(parent, startVertex, endVertex, map) {
       coordinates: [mid.lng, mid.lat]
     }
   };
+  delete midpoint.properties.id;
+  return midpoint;
 }
--- a/src/lib/create_midpoint.js
+++ b/src/lib/create_midpoint.js
@@ -1,4 +1,4 @@
-import * as Constants from "../constants";
+import * as Constants from '../constants';
 
 /**
  * Returns GeoJSON for a Point representing the
--- a/src/lib/create_supplementary_points.js
+++ b/src/lib/create_supplementary_points.js
@@ -4,13 +4,12 @@ import * as Constants from '../constants';
 
 function createSupplementaryPoints(geojson, options = {}, basePath = null) {
   const { type, coordinates } = geojson.geometry;
-  const featureId = geojson.properties && geojson.properties.id;
 
   let supplementaryPoints = [];
 
   if (type === Constants.geojsonTypes.POINT) {
     // For points, just create a vertex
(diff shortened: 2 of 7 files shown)


JamesLMilner:
@asheemmamoowala thanks for the review, much appreciated! What would be the next steps to get this merged? I can see #944 is also in a similar state. Would be amazing to get these two approved PRs merged if possible as they'd be very beneficial for us and the community at large :smile:

JamesLMilner:
@kkaefer thanks for merging the other PR. I saw you merged master into this PR, so thanks for keeping this up to date too. Any update on the steps that need to happen to get this merged? :)

usb248:
+1

awulkan:
Has this been put on hold again? There has been pull requests for this since 2018 and they always get stuck. What is the plan, are you waiting for something specific before it can be merged and released?

caldwellc:
+1

src/lib/create_midpoint.js
```diff
@@ -1,27 +1,40 @@
...
-  const mid = map.unproject([ (ptA.x + ptB.x) / 2, (ptA.y + ptB.y) / 2 ]);
+  const ptA = map.project([startCoord[0], startCoord[1]]);
+  const ptB = map.project([endCoord[0], endCoord[1]]);
+  const mid = map.unproject([(ptA.x + ptB.x) / 2, (ptA.y + ptB.y) / 2]);
 
-  return {
+  const midpoint = {
     type: Constants.geojsonTypes.FEATURE,
     properties: {
+      ...parent.properties,
```
adrianababakanian, line 35:
```suggestion

```

adrianababakanian, line 35: It's worth noting that, per [MDN's browser compatibility docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), IE does not support spread syntax.

JamesLMilner: That's a good spot, I think a `for ... in` loop might cover it for IE support

JamesLMilner:
Hi @adrianababakanian, thanks for your follow up, really happy to hear you had a chance to test it locally. 

Performance isn't something I had considered as I'd mostly taken @drewbo code and brought it in line with `main`. I can see why that might be problematic however as it grows.

I am pondering about your suggestion about looking up the parent properties using the ID, and the one thing I'm not sure about is if that would be possible using the Mapbox GL style specification? Maybe an alternative solution is to keep the parent as a reference to the parent object then you could do a `match` lookup on the `parent`  properties? That might have added knock on effects and added complexities however. I guess I am trying to determine how it would be possible to style the vertices based on the parent properties using only the parent ID, if that makes sense?

JamesLMilner:
@adrianababakanian I experimented with using a `parentProperties` property on the vertex/midpoints which should allow for you to write a style like this:

```
{
    id: 'gl-draw-polygon-and-line-midpoint-halo-parent-prop',
    type: 'circle',
    filter: [
        'all',
        ['match', ['geometry-type'], 'Point', true, false],
        ['match', ['get', 'meta'], 'midpoint', true, false],
        ['match', ['get', 'parentProperties', ['get', 'someParentProp'], true, false]
    ],
    paint: {
        'circle-radius': 5,
        'circle-color': '#FFF'
    }
}
```

You can check it here:
https://github.com/mapbox/mapbox-gl-draw/compare/main...JamesLMilner:846-midpoint-user-props-parent-props

Perhaps this would be a more performant approach as we don't copy properties in this instance?

usb248:
> 
> 
> @adrianababakanian I experimented with using a `parentProperties` property on the vertex/midpoints which should allow for you to write a style like this:
> 
> ```js
> {
>     id: 'gl-draw-polygon-and-line-midpoint-halo-parent-prop',
>     type: 'circle',
>     filter: [
>         'all',
>         ['match', ['geometry-type'], 'Point', true, false],
>         ['match', ['get', 'meta'], 'midpoint', true, false],
>         ['match', ['get', 'parentProperties', ['get', 'someParentProp'], true, false]
>     ],
>     paint: {
>         'circle-radius': 5,
>         'circle-color': '#FFF'
>     }
> }
> ```
> 
> You can check it here:
> [main...JamesLMilner:846-midpoint-user-props-parent-props](https://github.com/mapbox/mapbox-gl-draw/compare/main...JamesLMilner:846-midpoint-user-props-parent-props)
> 
> Perhaps this would be a more performant approach as we don't copy properties in this instance?

This approach doesn't allow us to do dynamic style. Mapbox GL style specification doesn't allow to look/get properties on another geojson. It means that you can get user's properties in a vextex from (a parent) feature (linestring) for example.

Zirak:
Would it be feasible to use `Object.create` or similar to leverage inheritance, instead of copying?

adrianababakanian:
Thanks for looking into this further @JamesLMilner . I agree that trying to make this work with only a reference to the parent ID adds more complexities than just copying the parent's `properties`. @usb248 brings up a good point that copying `properties` would be necessary to support fully dynamic styling. 

In my opinion, introducing the Feature State API to GL Draw will be beneficial in the long term. Starting to use feature state will involve a larger refactor of GL Draw as a whole, but will bring many additional benefits beyond just the feature in this PR. I opened https://github.com/mapbox/mapbox-gl-draw/issues/994 to start a discussion around a few of these benefits. Using feature state would enable giving child features (i.e. midpoints and vertices) the same `id` as the parent feature for styling purposes.

jaybo:
Is it correct to assume this same issue exists with **'Points'** also?  I'm trying to use:
`layout: {
    ...
     "text-field": ['get', 'name'],
    ...
},`
or alternatively:
`layout: {
    ...
     "text-field": "{name}",
    ...
},`
But it appears `properties.name` is not available to the style.  Is this a generic issue with Draw, that properties are not available for styling?

jbeuckm:
Can we merge this and add a warning in the md that adding a larger number of properties could introduce performance problems, especially with complicated shapes that render many midpoints and vertices? Not being able to style midpoints and vertices is holding us back.

chenlevy4:
@adrianababakanian  Can you please merge this PR?

ThomasAribart:
@adrianababakanian @asheemmamoowala Need this PR for my project as well 🙏 Could you merge and release it please ?

igamidanes:
This would be amazing for us as well

PR #969, opened 2020-03-09
Optimize SVG icons

manuelkasper:
Removed Inkscape etc. bloat using SVGO, and applied URL encoding (with https://yoksel.github.io/url-encoder/) instead of Base64.

CSS file size shrinks from about 32.1 to 5.5 KB uncompressed, or from about 8.8 to 1.5 KB gzipped.


asheemmamoowala:
Closing as duplicate of #1038

PR #972, opened 2020-03-20
Don't call onDrag when multiple fingers are touching the map

### LINKED ISSUE #962: When plugin enable there is a significant delay when initiating zoom with 2 finger pinch on mobile devices
**mapbox-gl-js version**: latest
**mapbox-gl-draw version**: latest

### Steps to Trigger Behavior

 1. open a mobile browser and go here: https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/
 2. zoom using 2 fingers pinching outwards 
 3. notice a delay on the initial action of zooming

if you hold your 2 fingers there for a short period it works when you move them but if you touch the screen with 2 fingers and immediately attempt to zoom using this method there is a delay. This delay does not seem to exist on desktop computers when doing the same gesture on a trackpad

### Expected Behavior
Smooth delay-free zoom

### Actual Behavior
Zoom delayed at initial 2 finger touch with immediate zoom

trygveaa:
Calling onDrag when multiple fingers are touching the map causes a delay
when you start pinch zooming. This is because the pinch is interpreted
as a tap in the beginning, which causes stopPropagation to be called.

I think dragging is an action you only do with one finger, so therefore
I think not calling onDrag is the correct solution to this.

I'm not sure how to test this, as I haven't figured out how to check if events are called. Are there any existing tests doing such checks?

Fixes #962

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -101,7 +101,10 @@ export default function(ctx) {
     }
 
     currentMode.touchmove(event);
-    return events.touchdrag(event);
+
+    if (event.points.length === 1) {
+      events.touchdrag(event);
+    }
   };
 
   events.touchend = function(event) {


underbluewaters:
I've used this approach in my own fork to avoid wrecking a shape that's in direct_select mode while pinch-zooming on a mobile device. I'm finding editing on mobile to be pretty impossible without that change.

PR #975, opened 2020-03-20
Allow deleting the whole feature from direct_select

trygveaa:
When no vertices are selected, trash deletes the whole feature.


neodescis:
I think this change would be great. It's not always immediately obvious to the user whether they're in direct or simple select mode. If no vertices are selected, it's pretty strange to just do nothing when hitting trash/delete.

PR #976, opened 2020-03-20
Set the compiled file in dist as the main file

trygveaa:
This lets people just import @mapbox/mapbox-gl-draw instead of having to
import @mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.js. It's standard
practise to set the compiled file in main, and most other packages
including mapbox-gl does so.


stevage:
It looks like the change proposed by this PR was made by #1032, so this can be closed now.

PR #977, opened 2020-03-20
Include shims for the fs and path dependencies

### LINKED ISSUE #970: Built file in master requires fs
**mapbox-gl-js version**: 1.8.1
**mapbox-gl-draw version**: master

### Steps to Trigger Behavior

 1. Clone this repo and run `yarn && yarn build-min`
 2. Try to import `dist/mapbox-gl-draw.js` in a project

### Expected Behavior

It should work, without depending on `fs`.

### Actual Behavior

I can't use it because it depends on `fs`. That is, I'm getting:

```
ERROR in ./node_modules/@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.js
Module not found: Error: Can't resolve 'fs' in '/home/trygve/dev/my-project/front/node_modules/@mapbox/mapbox-gl-draw/dist'
```

If I try to do the same after checking out `v1.1.2` it works fine.

trygveaa:
Including fs and path in externals is not sufficient when this is used
in a browser. That will just make rollup leave the requires calls, and
when you try to include this library in a webpack project webpack will
complain that fs is missing.

By providing shims for fs and path, those will be used instead of
require calls, so that fixes the issue. Since no code paths in this
library triggers the code that requires fs or path it's safe to just
shim them out with an empty object.

Fixes #970

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "opener": "^1.4.1",
     "rollup": "^1.27.13",
     "rollup-plugin-commonjs": "^10.1.0",
+    "rollup-plugin-shim": "^1.0.0",
     "rollup-plugin-terser": "^5.1.3",
     "sinon": "^7.5.0",
     "st": "^2.0.0",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -7,6 +7,7 @@ import buble from '@rollup/plugin-buble';
 import {terser} from 'rollup-plugin-terser';
 import resolve from '@rollup/plugin-node-resolve';
 import commonjs from 'rollup-plugin-commonjs';
+import shim from 'rollup-plugin-shim';
 
 export default {
   input: ['index.js'],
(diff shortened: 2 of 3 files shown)


trygveaa:
As an aside, I see that the rollup config makes rollup bundle all of the dependencies into the dist file. This seems like a poor thing to do?

By the way, this happens in master too, it's not because of the change in this PR.

trygveaa:
I do think that the `geojsonhint` dependency should probably be dropped instead though. The readme states that development of it has paused, and `jsonlint-lines` which it depends on isn't developed anymore either and is intended for CLI usage.

Helpful error messages like this provides might be useful for development, but I don't really see the point of it when this library is used in production. It's not like the user calls `add` themselves. 

Furthermore, Mapbox itself doesn't give such detailed error messages when you try to add a source with invalid data. Why should this library do it different than Mapbox?

stepankuzmin:
Thanks for the contribution, @trygveaa. Draw no longer depends on geojsonhint, so I will close this.

PR #978, opened 2020-03-20
When clicking an inactive feature in direct_select, select it

trygveaa:
With this, you don't have to click twice to select another feature when
you are in direct_select.

The mode will still be changed to simple_select. So if you want to
activate direct_select for another feature, you have to click it twice,
but that's still one less click than before.


usb248:
Good ! But cursor pointer on inactive feature is missing

trygveaa:
@usb248: Ah, right. I fixed it now.

test/direct_select.test.js
```diff
@@ -273,6 +273,22 @@ test('direct_select', (t) => {
...
+      featureId: lineId
+    });
+    const clickAt = getGeoJSON('polygon').geometry.coordinates[0][0];
+    afterNextRender(() => {
+      click(map, makeMouseEvent(clickAt[0], clickAt[1]));
+      afterNextRender(() => {
+        t.equal(Draw.getSelectedIds().indexOf(polygonId) !== -1, true, 'polygon is now selected');
+        cleanUp(() => st.end());
+      });
+    });
```
adrianababakanian, line 289: Since a selection change is occurring, it is worth checking that the correct event is fired.
```suggestion
    afterNextRender(() => {
      click(map, makeMouseEvent(clickAt[0], clickAt[1]));
      map.fire.resetHistory();
      afterNextRender(() => {
        const args = getFireArgs().filter(arg => arg[0] === 'draw.selectionchange');
        t.equal(args.length, 1, 'should have one and only one selectionchange event');
        t.equal(Draw.getSelectedIds().indexOf(polygonId) !== -1, true, 'polygon is now selected');
        cleanUp(() => st.end());
      });
    });
```

trygveaa: Good idea, applied.

PR #983, opened 2020-04-29
Adds insertBeforeLayerId option

kukiel:
Using this option user can specify id of the layer before all MapboxDraw layers should be inserted.

Happy to improve naming/tests if needed.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -36,6 +36,7 @@ All of the following options are optional.
 - `modes`, Object: over ride the default modes with your own. `MapboxDraw.modes` can be used to see the default values. More information on custom modes [can be found here](https://github.com/mapbox/mapbox-gl-draw/blob/master/docs/MODES.md).
 - `defaultMode`, String (default: `'simple_select'`): the mode (from `modes`) that user will first land in.
 - `userProperties`, boolean (default: `false`): properties of a feature will also be available for styling and prefixed with `user_`, e.g., `['==', 'user_custom_label', 'Example']`
+- `insertBeforeLayerId`, string (default: `undefined`): The ID of an existing layer to insert MapboxDraw draw layers before. If this argument is not specified, layers will be appended to the end of the layers array - [more here](https://docs.mapbox.com/mapbox-gl-js/api/#map#addlayer). 
 
 ## Modes
 
--- a/test/layers_insert.test.js
+++ b/test/layers_insert.test.js
@@ -1,0 +1,32 @@
+/* eslint no-shadow:[0] */
+import test from "tape";
+import MapboxDraw from '../';
+import createMap from "./utils/create_map";
+import spy from "sinon/lib/sinon/spy";
+
+test("Layers insert test", (t) => {
+  t.test("add map layer before insertBeforeLayerId", (t) => {
+    const layerId = "test";
+    const map = createMap();
+    const addLayerSpy = spy(map, "addLayer");
+    const opts = {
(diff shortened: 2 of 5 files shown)


JamesLMilner:
I can see the use cases here, especially for those using Mapbox GL Draw in more complex applications. Seems like a nice addition to the API to me.

kukiel:
> I can see the use cases here, especially for those using Mapbox GL Draw in more complex applications. Seems like a nice addition to the API to me.

@JamesLMilner yes, that was a killer for me when having ~30layer and some of them need to be above the drawings.

stevage:
FWIW there is a fairly easy workaround - you can simply call `map.moveLayer()` on all those layers after draw is initiated.

docs/API.md
```diff
@@ -36,6 +36,7 @@ All of the following options are optional.
 - `modes`, Object: over ride the default modes with your own. `MapboxDraw.modes` can be used to see the default values. More information on custom modes [can be found here](https://github.com/mapbox/mapbox-gl-draw/blob/master/docs/MODES.md).
 - `defaultMode`, String (default: `'simple_select'`): the mode (from `modes`) that user will first land in.
 - `userProperties`, boolean (default: `false`): properties of a feature will also be available for styling and prefixed with `user_`, e.g., `['==', 'user_custom_label', 'Example']`
+- `insertBeforeLayerId`, string (default: `undefined`): The ID of an existing layer to insert MapboxDraw draw layers before. If this argument is not specified, layers will be appended to the end of the layers array - [more here](https://docs.mapbox.com/mapbox-gl-js/api/#map#addlayer).
```
adrianababakanian, line 39: For parity with GL JS methods such as [`Map#addLayer`](https://docs.mapbox.com/mapbox-gl-js/api/map/#map#addlayer) and [`Map#moveLayer`](https://docs.mapbox.com/mapbox-gl-js/api/map/#map#movelayer), it might be clearer + conciser to name the option `beforeId` rather than `insertBeforeLayerId`.

test/layers_insert.test.js
```diff
@@ -0,0 +1,32 @@
+/* eslint no-shadow:[0] */
+import test from "tape";
+import MapboxDraw from '../';
+import createMap from "./utils/create_map";
+import spy from "sinon/lib/sinon/spy";
+
+test("Layers insert test", (t) => {
```
adrianababakanian, line 7: It would be beneficial to test that the drawing functionality (i.e. adding + displaying vertices) still works as expected when drawing "beneath" a layer specified with `insertBeforeLayerId` in a region where said layer has features defined. That is, suppose `insertBeforeLayerId` is set to `'layer'`, and `'layer'` has a polygon feature covering a geographic region where the user is drawing.

Adding a test to clarify the intended behavior of selecting a drawn layer being "covered" by a feature in `'layer'` would also be helpful.

kukiel:
@stevage 
I was not aware of that function in mapbox API! I think it would solve the use case without any additional changes in draw library code. Thanks a lot.

@adrianababakanian 
Seeing the comment from @stevage I will probably refactor my code at some point and use `moveLayer` to avoid extending draw library API when its actually not necessary.

Thanks guys.

ShareQiu1994:
I've had this problem, too. It can be used as mentioned above map.moveLayer (id, beforeid) adjust the layer order, but I can't find it mapbox.gl.draw Many methods have been tried to set the layer id. in the end, this problem can only be solved after the new mapboxdraw is placed.

Issue #684, opened 2017-08-31
map.on('draw.update', callback) does not fire consistently


Plysepter:
**mapbox-gl-js version**: v0.39.1, v0.38.0, v0.37.0
**mapbox-gl-draw version**: v1.0.0, 0.19.0

### Steps to Trigger Behavior

 1. Open Plunkr: https://plnkr.co/edit/Rn5AZKlDrdBBAQmdiRy2?p=preview
 2. Open Dev Tools -> Console
 3. Draw a polygon
 4. Update the polygon by moving a point
 5. Repeat step 4 and keep track of the number of times you have moved a point. After a few changes you should notice that the count variable does not represent the number of changes actually made.

### Expected Behavior
The `draw.update` event callback should fire each time a point is moved to a new position. Which in turn would increment the count variable by one.

### Actual Behavior
The `draw.update` event callback is not being fired consistently.

IAfanasov:
Opened plunker and received error:

api.mapbox.com/styles/v1/mapbox/streets-v9?access_token=pk.eyJ1Ijoic2t5d2F0Y2hhcHBzIiwiYSI6ImNpa3B4aHp6OTE1OWV0bm02aWQwOTd2OG8ifQ.Ic8z13iLo4r39I01ckGvZA Failed to load resource: the server responded with a status of 401 (Unauthorized)

Plysepter:
Updated Plunker with new key. Sorry about that

IAfanasov:
moved different points about 20 times. all fired. may you clarify the scenario? may be screencast?

Plysepter:
I have provided a screencast to better demonstrate the issue. You can skip to about 19 seconds in for the updating to start.

Here are a list of time codes for missed events:
0:22 - Between console log 3 & 4
0:26 - Between console log 5 & 6
0:28 - Between console log 6 & 7
0:36 - After console log 14

I hope that the screen capture better illustrates the issue. I am not sure how to provide more detail other than to share the screen capture and note that while I am able to easily reproduce the issue, there is no specific pattern observed for when the event doesn't fire.

Please let me know if the screen capture helped or if I can be of any more help in tracking down this bug. I appreciate the help

[UpdateNotFiring.mp4.zip](https://github.com/mapbox/mapbox-gl-draw/files/1352345/UpdateNotFiring.mp4.zip)

IAfanasov:
[...]

PR #1008, opened 2020-08-27
Fix 'simple_select' mode lacking 'touchend' event handling (#684)

corinv:
Fixes (#684). 

The `draw.update` event would not be fired on moving a geometry using a touch device (or emulated touch, like Chrome's device mode) as there was no handling for a `touchend` event to fire said event in the `simple_select` mode.


gbuckingham89:
Would love to see this merged - it seems to fix an issue we're having on touch devices (iPad) where dragging polygon in `simple_touch` mode never ends until a tap is made (on the polygon or the map) - meaning you can't touch and rag the polygon, then immediately touch to pan the map.

mourner (approved):

mourner:
Sorry for getting back so late on this, and thanks for the contribution!

PR #1016, opened 2020-09-30
Update guide for Developing and testing

baspeti:
...so that novice users can try it too. For example I thought `yarn start` is also serving the app on port 8867, but it turned out that it did not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Install dependencies, build the source files and crank up a server via:
 ```
 git clone git@github.com:mapbox/mapbox-gl-draw.git
 yarn install
+node server.js
 yarn start & open "http://localhost:9967/debug/?access_token=<token>"
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,6 +83,7 @@ Install dependencies, build the source files and crank up a server via:
 ```
 git clone git@github.com:mapbox/mapbox-gl-draw.git
 yarn install
+node server.js
```
rreusser, line 86: Maybe instead we can add `"start-server": "node server.js"` to `scripts` and then add `start-server` to the `start` script so that `yarn start` will do it all.

rreusser (changes requested):

rreusser:
I think this can be closed in favor of #1040 so that you don't have to manually start the dev server. cc @arindam1993

PR #1017, opened 2020-10-01
Fix boxZoom bug

### LINKED ISSUE #571: Draw removes boxZoom but never restores it on removal
During setup Draw removes boxZoom (to be able to provide shift-box-selections instead):

https://github.com/mapbox/mapbox-gl-draw/blob/78fddb17520f2869902b2e6dbf93c1bf888cf421/src/setup.js#L39

But when draw is removed it never tries to enable it again.

kibala145:
Closes #571


kibala145:
anyone please

stepankuzmin (approved):

PR #1052, opened 2021-05-19
Remove GeoJSON validation

### LINKED ISSUE #1020: NPM audit reports vulerabilities
**mapbox-gl-js version**: 1.12.0 (but doesn't really matter)

### Steps to Trigger Behavior

 1. Init an npm repo using "npm init" and pressing "Enter" for all prompts
 2. Run "npm i @mapbox/mapbox-gl-draw"
 3. Run "npm audit"

### Expected Behavior
0 vulnerabilities reported.

### Actual Behavior
2 vulnerabilities reported:
mapbox-gl-draw-vulnerabilities

### Additional info
Both vulnerabilities are caused by the minimist package using 1.2.0 version in which there is a low vulnerability. It is fixed in the newest version of minimist.

To fix the "@mapbox/mapbox-gl-draw > @mapbox/geojson-extent > @mapbox/geojson-coords > geojson-flatten > minimist" all you need to do is to bump the @mapbox/geojson-extent version to **1.0.0** and every library down the path is already updated to version without the vulnerability.

For the "@mapbox/mapbox-gl-draw > @mapbox/geojsonhint > minimist" path, @mapbox/geojsonhint has already [updated](https://github.com/mapbox/geojsonhint/pull/79) the minimist version to a non-vulnerable version, but the change isn't released yet. There is an [issue](https://github.com/mapbox/geojsonhint/issues/78) about that, but there is no activity since May of this year.

I know that this vulnerability can't really be exploited from the mapbox-draw-gl package, but my company has a blanket "no vulnerabilities" policy, so if you can update those packages, it will be greatly appreciated.

### LINKED ISSUE #1049: npm vulnerability
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Arbitrary Code Execution                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ underscore                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=1.12.1                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @mapbox/mapbox-gl-draw                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @mapbox/mapbox-gl-draw > @mapbox/geojsonhint >               │
│               │ jsonlint-lines > nomnom > underscore                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1674                            │
└───────────────┴──────────────────────────────────────────────────────────────┘

### LINKED ISSUE #1051: Remove dependency on @mapbox/geojsonhint
**mapbox-gl-js version**: n/a
**mapbox-gl-draw version**: all

### Issue

Currently this package (mapbox-gl-draw) depends on the `@mapbox/geojsonhint` package, which has been marked as deprecated in npm: https://www.npmjs.com/package/@mapbox/geojsonhint

This also introduces some security vulnerabilities via the dependency chain of `@mapbox/geojsonhint`. As far as I can tell, this package is only used to perform validation of geojson via the `add` API: https://github.com/mapbox/mapbox-gl-draw/blob/main/src/api.js#L77-L80

The recommendation from `@mapbox/geojsonhint` is to perform validation some other way; Removing the deep integration here would only have a minor affect on usability of this module, as only the `add` method did validation of the geojson, and all others didn't.

If you would like a PR to remove this dependency, please let me know and I'll prepare one.

mourner:
Closes #1051, closes #1020, closes #1049. Upgrades some dependencies to pass security audit, and removes GeoJSON validation which wasn't very useful for its added size, was applied inconsistently, and relied on an unmaintained library with highly vulnerable transitive deps. Removing it is technically not too breaking since the code that worked before will continue working after the upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,12 +87,12 @@
     "unassertify": "^2.0.3"
   },
   "dependencies": {
-    "@mapbox/geojson-area": "^0.2.1",
-    "@mapbox/geojson-extent": "^0.3.2",
-    "@mapbox/geojson-normalize": "0.0.1",
-    "@mapbox/point-geometry": "0.1.0",
+    "@mapbox/geojson-area": "^0.2.2",
+    "@mapbox/geojson-extent": "^1.0.0",
+    "@mapbox/geojson-normalize": "^0.0.1",
+    "@mapbox/point-geometry": "^0.1.0",
     "hat": "0.0.3",
-    "lodash.isequal": "^4.2.0",
-    "xtend": "^4.0.1"
+    "lodash.isequal": "^4.5.0",
+    "xtend": "^4.0.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -86,9 +86,6 @@
     "uglify-js": "^3.0.22",
     "unassertify": "^2.0.3"
   },
-  "peerDependencies": {
-    "mapbox-gl": ">=0.27.0"
-  },
   "dependencies": {
     "@mapbox/geojson-area": "^0.2.1",
     "@mapbox/geojson-extent": "^0.3.2",
--- a/package.json
+++ b/package.json
@@ -90,7 +90,6 @@
     "@mapbox/geojson-area": "^0.2.1",
     "@mapbox/geojson-extent": "^0.3.2",
     "@mapbox/geojson-normalize": "0.0.1",
-    "@mapbox/geojsonhint": "3.0.0",
     "@mapbox/point-geometry": "0.1.0",
     "hat": "0.0.3",
     "lodash.isequal": "^4.2.0",
--- a/src/api.js
+++ b/src/api.js
@@ -3,7 +3,6 @@ import normalize from '@mapbox/geojson-normalize';
 import hat from 'hat';
 import featuresAt from './lib/features_at';
 import stringSetsAreEqual from './lib/string_sets_are_equal';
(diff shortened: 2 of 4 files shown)


rreusser (approved):
This makes sense to me. Do I understand correctly that users would then be responsible for passing valid geojson? Should we add a note to [the docs for `add`](https://github.com/mapbox/mapbox-gl-draw/blob/main/docs/API.md#addgeojson-object--arraystring) which conveys this?

mourner:
I think it's fine to not mention — the old validation behavior was never mentioned anyway, and requirement to pass valid GeoJSON is implied in all APIs like this

davidbeers:
This is great. @mourner do you have a rough idea when there might be a release with this fix? From reading comments it seems like it would make a very welcome minor release all on its own even if there aren't other features or fixes ready to release.

mourner:
@davidbeers just released!

murdocha:
Thank you!
this is much better than the work-around I I had found to reference a Github commit by hash in package.json:
```
    "@mapbox/mapbox-gl-draw": "git+https://github.com/mapbox/mapbox-gl-draw.git#c12fb64b90d00e877e94b12b60bc3b80e42924dd",
```

PR #1066, opened 2021-08-20
Remove geojson-flatten dependency

stevage:
This dependency on geojson-flatten (which itself has several more) is easily replaced.

It also turns out many of the tests to `constrainFeatureMovement()` were supplying faulty Polygon coordinates.

(Motivation: this was causing me some kind of spurious packaging errors when making other changes, the classic Webpack issue with `require()` and `.default` etc etc)


stepankuzmin (approved):
Hey @stevage,

Thanks for the contribution, and sorry for the delayed response. This change also saves 8K in the resulting bundle 👍

PR #1068, opened 2021-09-07
Add Bezier Curve mode

Jeff-Numix:
A mapbox gl draw mode to draw and edit bezier curves.


SnailBones:
Thanks for your contribution @Jeff-Numix! This is a really cool feature.

Issue #1070, opened 2021-10-05
Can't return to original coords after dragging


oodavid:
**mapbox-gl-js version**: 2.3.1
**mapbox-gl-draw version**: 1.2.2

### Steps to Trigger Behavior

Start dragging a polygon and attempt to return it to the original coordinates

https://www.loom.com/share/9ac3e72134074d049211a3027276d223

### Expected Behavior

I should be able to drag the polygon back to the original position

### Actual Behavior

There is a "deadzone" where I cannot place the polygin

It may affect other drags (point, line etc.)

oodavid:
...yup, it applies to all draggable elements, including the polygon line handles

PR #1075, opened 2021-10-28
Allow drag events to return to their original coords

oodavid:
Fix for #1070

Simplified the is_click and is_tap tests


oodavid:
See PR #1076 instead

PR #1079, opened 2021-11-05
fix: Change how events.keydown checks classList for mapboxgl-canvas 

thaddmt:
### Overview
- Originally `events.keydown` only checked the first value in the `classList` to see that it matches `mapboxgl-canvas`
- Updated so that it checks the full `classList` for the value `mapboxgl-canvas`
  - This enables usage with a classList that starts with `maplibregl-canvas`  for instance
- **NOTE - Contains vs Includes**
  - `DOMTokenList` only supports `contains`
  - In our tests we create synthetic events with an array for a classList which only supports `includes`
  - Couldn't create a `DOMTokenList` without creating an element so went with this approach

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -127,7 +127,9 @@ export default function(ctx) {
   const isKeyModeValid = code => !(code === 8 || code === 46 || (code >= 48 && code <= 57));
 
   events.keydown = function(event) {
-    if ((event.srcElement || event.target).classList[0] !== 'mapboxgl-canvas') return; // we only handle events on the map
+    const classList = (event.srcElement || event.target).classList;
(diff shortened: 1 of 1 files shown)


TreTuna (approved):

HarelM (approved):

robinsummerhill:
v3.0.0 of maplibre-gl-js removed legacy `mapboxgl-*` classes in this PR https://github.com/maplibre/maplibre-gl-js/pull/1575

stepankuzmin:
This was closed in https://github.com/mapbox/mapbox-gl-draw/pull/1165. Now available in `v1.4.2` https://github.com/mapbox/mapbox-gl-draw/releases/tag/v1.4.2

PR #1093, opened 2022-02-10
fix: syntax error in example styles

maxnowack:


mourner:
Good catch, thanks! Already merged in #1111 that fixed the same issue so closing...

PR #1099, opened 2022-03-15
Update style examples to match API.md

carderne:


stepankuzmin:
Good catch, thanks, @carderne!

Issue #967, opened 2020-02-21
Can't set multiple selected coordinates when switching to Direct Select mode


DannyDelott:
I'm writing a custom mode that lets a user box select over a Line feature to select multiple vertices, then pass them off to `direct_select` so they can be moved around or deleted. 

This is convenient when the Line feature is a windy road or otherwise contains lots of vertices where shift + clicking each one would be a hassle.

The problem I'm running into is that Direct Select only accepts a single `coordPath` in onSetup, which means we can only pass a single vertex off to Direct Select mode.

https://github.com/mapbox/mapbox-gl-draw/blob/e3edcc29a5fffdcdc416cbad4dfe118f38f0c845/src/modes/direct_select.js#L131

To make this work, I've forked Direct Select mode and tweaked the above line of code so that the caller is responsible for passing a `coordPath` array, instead of expecting onSetup to wrap it:

```js
 selectedCoordPaths: opts.coordPath ? opts.coordPath : [] 
```

This gives me my expected result.

![box-select-mapbox](https://user-images.githubusercontent.com/4524175/74993684-a1d44e00-5400-11ea-86e7-00d2aa81991c.gif)

If it makes sense, I'm happy to push up a PR with this small change. Or maybe there's a more idiomatic way to do this that I'm not seeing :)

DuyErasA:
Hi @DannyDelott ,

Any chance you could share your custom mode?

Best

xarowls:
Hi @DannyDelott , 
Thanks for sharing. Would love to understand how you captured the vertices. 
Any chance you could share the custom mode with an example?

Cheers!

alt-ctrl-dev:
Hi @DannyDelott, thanks for sharing. 

I've made a pull request. Hope this helps
https://github.com/mapbox/mapbox-gl-draw/pull/1115

zbyte64:
For those curious of @DannyDelott changes to support this feature, you can view the diff we used to goto production to support this feature here: https://github.com/mapbox/mapbox-gl-draw/compare/main...CalthorpeAnalytics:mapbox-gl-draw:master#diff-c5de164fc4c2e06e1053418f2c5127ba4fafe0ad283a9349412ddefb83a07c38R308

PR #1115, opened 2022-07-21
Allowing multiple coordinate paths to be passed

alt-ctrl-dev:
This can help support custom draw mode tools for multi point option


alt-ctrl-dev:
Would help close #967

PR #1141, opened 2023-01-09
Keep midpoint on the line 

zhangchn:
With the average of Mercator coordinates of the two endpoints, the middle points were drawn off from the line.

By project/unproject the Mercator latitudes to and from screen space, the screen location for midpoints would be fixed onto the line.


stepankuzmin:
Hi @zhangchn, 

Thanks for the PR! Could you please review the failing tests in the CI?

zhangchn:
IMHO, my patch changed the way midpoints were located. Test cases which assume the midpoints were located the old way failed.

PR #1146, opened 2023-01-18
fixes #1019 

### LINKED ISSUE #1019: Unable to preventDefault inside passive event listener invocation.
Hello,

**mapbox-gl-js version**: 1.12

### Steps to Trigger Behavior

 1. Add Mapbox Draw control to mapbox map.
 2. Open in browser and switch to touch device.
 3. Click on map and check console 

Just open official web and switch to device toolbar in developer console. Touch a map and check console.
https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/

### Expected Behavior
Should not generate that issue. If any mapbox layers was added, they are not reacting to click/touch event. Only drawing tools and features which was drawed are "alive".
### Actual Behavior
After clicking on map when mapbox draw was added as control, mapbox draw generates error.  
Error text : 
" Unable to preventDefault inside passive event listener invocation."

danielsippel:
fixes:
- popups not opening by click event on mobile browsers if mapbox-draw activated
- chrome mobile console error "Unable to preventDefault inside passive event listener invocation."

More information about the issue & solution:
- https://chromestatus.com/feature/5093566007214080
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "build": "rollup -c",
     "build-min": "rollup -c --environment MINIFY:true",
     "prepublishOnly": "run-s build build-min",
+    "prepare": "run-s build build-min",
     "watch": "rollup -c --watch",
     "watch-bench": "rollup -c bench/rollup.config.js --watch",
     "start-server": "st --no-cache -H 0.0.0.0 --port 9967 --index index.html .",
--- a/src/events.js
+++ b/src/events.js
@@ -78,9 +78,6 @@ export default function(ctx) {
   };
 
   events.touchstart = function(event) {
-    // Prevent emulated mouse events because we will fully handle the touch here.
-    // This does not stop the touch events from propogating to mapbox though.
-    event.originalEvent.preventDefault();
     if (!ctx.options.touchEnabled) {
       return;
     }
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
     "build": "rollup -c",
     "build-min": "rollup -c --environment MINIFY:true",
     "prepublishOnly": "run-s build build-min",
+    "prepare": "run-s build build-min",
```
stepankuzmin, line 23: Hi @danielsippel! Sorry for the delayed review. What's the purpose of adding the `prepare` script here?

danielsippel: Hey @stepankuzmin !
When others npm install the mapbox-gl-draw dependency from e.g. my fork on github, the prepare script is run automatically by npm.. So the JS files are provided in /node_modules/dist/ automatically. Without the prepare script, the dist directory is empty after install. So forks can't be used without further manual build step.

stepankuzmin: So you are currently using a fork with GitHub URL in your dependencies? We can probably safely drop this in the main repo, then.

danielsippel: Right and yes.

stepankuzmin: I see, thanks! Could you please drop it in the PR?

danielsippel:
Done.

As mentioned, now my fork is not working as a dependency any longer.. 😅
So I would be glad if you could merge it soon :-)

PR #1155, opened 2023-02-09
fix: trigger layer update if it has been moved

gynekolog:
fix https://github.com/mapbox/mapbox-gl-draw/issues/684


gynekolog:
I'm going to move the commits from the main branch and create a new pull request.

PR #1159, opened 2023-02-14
Release v1.4.1

stepankuzmin:
The patch release includes the revert of passive event listeners in https://github.com/mapbox/mapbox-gl-draw/pull/1158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
-## Changelog
+# Changelog
 
-# 1.4.0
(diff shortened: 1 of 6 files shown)


emilhem:
Hmm, seems like this version isn't released properly on `api.mapbox.com`. Previous version (`1.4.0`) works but not `1.4.1`.

stepankuzmin:
Hi @emilhem, 

Good catch, thanks! Should be fixed now.

PR #1163, opened 2023-02-28
fix: set up events correctly when instantiating api

### LINKED ISSUE #1007: Cannot read property 'getMode' of undefined
### Cannot read property 'getMode' of undefined at MapboxDraw.__webpack_exports__.default.api.getMode

**mapbox-gl-js version**: 1.12.0

```
  draw.current = new MapboxDraw({
            displayControlsDefault: false,
            styles: drawingStyle,
            modes:
            {
                ...MapboxDraw.modes,
                [MapConstants.drawingTypeOption.freehand]: FreeDraw,
                lots_of_points: LotsOfPointsMode,
            }
        });
  console.log('draw.current.getMode', draw.current.getMode(), MapboxDraw.modes)
  map.addControl(draw.current);
```

https://github.com/mapbox/mapbox-gl-draw/blob/main/src/setup.js
the 'ctx' object missing events

add this line to solve this problem:
```
ctx.events = events(ctx);
```

StevenLangbroek:
Closes #1007. Ran into this issue when trying to make an external button display active state depending on the current mode. I verified that it works locally, but am unsure about testing policy. If this needs more work, let me know in comments and I'll get on it.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -4,6 +4,7 @@ import hat from 'hat';
 import featuresAt from './lib/features_at';
 import stringSetsAreEqual from './lib/string_sets_are_equal';
 import * as Constants from './constants';
+import events from './events';
 import StringSet from './lib/string_set';
 
 import Polygon from './feature_types/polygon';
@@ -21,6 +22,7 @@ const featureTypes = {
 };
 
 export default function(ctx, api) {
+  ctx.events = events(ctx);
 
   api.modes = Constants.modes;
 


stepankuzmin:
Hi @StevenLangbroek,

Thanks for making a PR for that! Could you please also add a small unit test in `test/api.test.js` that would check the original issue is no longer reproducible? https://github.com/mapbox/mapbox-gl-draw/issues/1007

StevenLangbroek:
> Hi @StevenLangbroek,
> 
> Thanks for making a PR for that! Could you please also add a small unit test in `test/api.test.js` that would check the original issue is no longer reproducible? https://github.com/mapbox/mapbox-gl-draw/issues/1007

No problem at all! I'm just glad I found the issue, the way Nextjs reported the error to me made me question my sanity for a bit haha.

StevenLangbroek:
@stepankuzmin if you have any hints as to how to test this, if I remove my fix and just add a test for `.getMode()` at the top of the file, nothing breaks.

stepankuzmin:
@StevenLangbroek, that's weird. Are you able to reproduce it locally in the browser?

StevenLangbroek:
@stepankuzmin yeah absolutely, that's why I started digging through issues in the first place... I'll add some breakpoints and step through until I understand the issue better. I'll reach out if I'm still unsure how to replicate under test, ok?

stepankuzmin:
@StevenLangbroek We don't have a specific release timeline for the GL Draw, but this fix would likely be a part of a larger release.

stepankuzmin:
Hi @StevenLangbroek,

Could you please try and call the `runSetup` before the `setupAPI` and see if that helps with the issue?

https://github.com/mapbox/mapbox-gl-draw/blob/e57aafe28a5a964f40c784cb8e2417f1adc161b7/index.js#L13-L18

ivari:
Any news on releasing this?

PR #1164, opened 2023-03-02
feat(events): in all element classes, look for canvas className

### LINKED ISSUE #1135: Keyboard events are not handled when using maplibre-gl-js
Currently the keydown event handler is not working for maplibre. I realize maplibre may not be officially supported, but MapboxDraw mostly works with it. The problem is that the keydown handler currently looks for a CSS class on the canvas, in a specific order in classList:

```
if ((event.srcElement || event.target).classList[0] !== 'mapboxgl-canvas') return; // we only handle events on the map
```

taken from here: https://github.com/mapbox/mapbox-gl-draw/blob/main/src/events.js#L130

In maplibre, they've added an additional class in front of that one, so that the canvas class is: "maplibregl-canvas mapboxgl-canvas", which causes the check above to bail out of handling the events.

So, my question is, would you take a pull request that changes the above line of code to something along the lines of !classList.contains('mapboxgl-canvas') instead? Or it could even look for either of the two above classes. I'm not sure how long maplibre is planning to keep around the mapbox CSS classes.

maplibre/maplibre-gl-js#2006

**maplibre-gl-js version**: 2.4.0
**mapbox-gl-draw version**: 1.3.0

### Steps to Trigger Behavior

 1. Select existing drawn shape
 2. Hit delete on the keyboard

### Expected Behavior

Polygon is deleted

### Actual Behavior

Polygon is not deleted

gynekolog:
fix #1135 

But the tests fail and I can solve/fix them.

Maybe someone else should be more successful.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -127,7 +127,8 @@ export default function(ctx) {
   const isKeyModeValid = code => !(code === 8 || code === 46 || (code >= 48 && code <= 57));
 
   events.keydown = function(event) {
-    if ((event.srcElement || event.target).classList[0] !== 'mapboxgl-canvas') return; // we only handle events on the map
+    const isMapElement = (event.srcElement || event.target).classList.contains('mapboxgl-canvas');
+    if (!isMapElement) return; // we only handle events on the map
 
     if ((event.keyCode === 8 || event.keyCode === 46) && ctx.options.controls.trash) {
       event.preventDefault();


neodescis:
Tests are failing because the test events have mock elements with classList variables that do not define contains(). I've opened [another PR](https://github.com/mapbox/mapbox-gl-draw/pull/1165) that addresses the tests.

gynekolog:
@neodescis Thank you for the better pull request, I'm going to close this one.

Issue #1182, opened 2023-06-29
MapboxDraw doesn't work with MapLibre 3.x


neodescis:
As of MapLibre 3.x, the MapboxDraw plugin is no longer working. The issue can be seen in this official [MapLibre example](https://maplibre.org/maplibre-gl-js-docs/example/mapbox-gl-draw/).

This is because MapLibre no longer uses the Mapbox CSS classes that MapboxDraw is checking for, which has been discussed in [this (closed) MapLibre issue](https://github.com/maplibre/maplibre-gl-js/issues/2601). Obviously this is because of a change to MapLibre, but I think it would be great if MapboxDraw could be used in both Mapbox and MapLibre. A relatively easy option I see is to consolidate all of the CSS classes that MapboxDraw is using into its list of constants (many but not all of them are already there), and allow them to be updated either before or while instantiating a MapboxDraw instance. Thoughts on this?

HarelM:
Sounds like a good approach from my point of view.
@pessimistress what is your solution for react-map-gl for these kind of issues?
@stepankuzmin what do you think should be the approach?

stepankuzmin:
Hey everyone! Overriding classes in GL JS Draw sounds like a good addition. We already have constants in `src/constants.js`, so implementing overriding should be straightforward. Does anyone want to take a first stab at it?

neodescis:
I'll be happy to give it a go. I'm hoping to be able to work on it this week.

bk201lucifer:
I tried this and it's okay
![a](https://github.com/mapbox/mapbox-gl-draw/assets/63433257/845e227e-08f7-47d9-ad1d-31fac10d6b8f)

mstahv:
Note to people finding this issue: it looks like only the "display controls" are broken. I was able to hack together a PoC for my Vaadin integration by using a separate UI to trigger the draw mode (and using defaultMode constructor parameter to start drawing immediately): https://github.com/parttio/maplibre/commit/edb0a498d5c384a3a35e2b8fdcc408075860db4c

geronimoooo:
[...]

PR #1100, opened 2022-03-16
chore: export constants and lib through MapboxDraw

thaddmt:
### Overview
- Saw there was an issue for exposing API's to make it easier to write custom draw modes - https://github.com/mapbox/mapbox-gl-draw/issues/1056
  - Experiencing a similar issue here - https://github.com/aws-amplify/maplibre-gl-draw-circle
- Added everything in `Constants` and `Lib` to the `MapboxDraw` default export
  -  Considered making them separate named exports from the default but wanted to keep it in line with `modes` which is also added directly onto the `MapboxDraw` object

PR #1184, opened 2023-07-12
Expose `CANVAS` class in constants

neodescis:
The intent is ultimately to close https://github.com/mapbox/mapbox-gl-draw/issues/1182.

To begin, I've pulled in the one mapbox-specific CSS class in the JavaScript code not yet in the list of constants. Thanks to https://github.com/mapbox/mapbox-gl-draw/pull/1100, these can already be updated readily before instantiating MapboxDraw, e.g.:

```
MapboxDraw.constants.classes.CONTROL_BASE = "maplibregl-ctrl";
MapboxDraw.constants.classes.CONTROL_PREFIX = "maplibregl-ctrl-";
MapboxDraw.constants.classes.CONTROL_GROUP = "maplibregl-ctrl-group";
...
```

If using @types/mapbox__mapbox-gl-draw, the current typings complain about this as they are marked readonly there (needing a @ts-ignore to compensate), so it would probably be good to create a pull request there too.

This doesn't make the library functional with MapLibre out-of-the-box though. The CSS distributed with mapbox-draw depends on the original values of the class constants, as well as mapbox class names. I'm not entirely sure what we would want to do about that either, though. Any thoughts, @HarelM or @stepankuzmin?


HarelM:
To replay to what you wrote:
1. Changing the `constants` from being readonly to not readonly is needed.
2. The initialization when using MapLibre should be documented in the MapLibre draw example - this library is Mapbox first, and until this is changed, we need to document how to use it with MapLibre. I don't see a way around that unless we want to introduce here an option of some kind to let the library know which one is being used (i.e. have another constant classes object that this library will use in case of a parameter sent in the initialization of the plugin - i.e. `useMaplibre` or something). Which means that it supports both Mapbox and MapLibre out of the box with no additional "hacking".
4. In continue to the above, one option is to have the classes doubled - i.e. have both mapbox and maplibre in the css. Another option is to do the dynamic css part that is mapbox/maplibre specific in the code and not in the css, thus allowing the dynamic nature of it. Since we are not talking about a lot of classes I think this is a possible solution.
5. If 4 is not possible one could create a css and host it on gist or something to support mapbox draw "missing" MapLibre css. this however will probably break at one point or another.

My two cents anyway, but I'm not maintaining this lib and I don't know if the out of the box support for MapLibre is welcomed or not. I would say that there are a lot of people who are using MapLibre with this lib (I got a large number of tickets on it in MapLibre's repo) and I believe they will want to contribute to this lib as well instead of forking, I would also like to avoid forking if possible.
Note that react-map-gl is also supporting both MapLibre and Mapbox so we could potentially learn from them.
You can see here that the docs has both maplibre and mapbox reference:
https://visgl.github.io/react-map-gl/docs/api-reference/map

gowin20:
+1 to maintaining MapLibre support out of the box in this plugin. I use Mapbox GL Draw frequently in my MapLibre apps, and I know several others who do the same. I'd be happy to contribute to this issue as well in order to continue maintaining compatibility.

mike8161990:
Any movement on this? I'm seeing the same bug and would love if this fix was incorporated.

neodescis:
I have not had the time to continue working on this. The way I see it, there are two additional things to resolve:

1. Either we need to update the typings for Mapbox Draw to allow the "constants" to be settable, or we need some other way of making them configurable.
2. We need some way to have the CSS's class names use the constants. Maybe this means moving the CSS into the JavaScript code, so that it can be generated on the fly?

However, I do not think those necessarily need to be handled in this pull request, so perhaps this could be merged as-is?

stepankuzmin (approved):
Hey all! I'm going to merge this. Again, thanks for the contribution, @neodescis

PR #1194, opened 2023-08-16
Fix for layer not found error in featuresAt function

### LINKED ISSUE #1183: map.setStyle() followed by mouse event sometimes causes layer not found error
When map.setStyle() is called and a mousemove event fires, mapbox-gl-draw's event listener will run map.queryRenderedFeatures passing the layers from ctx.options.styles. However, this sometimes occurs before mapbox-gl-draw's map.on("data") event listener can add back in the missing style layers caused by the map.setStyle() function, the following error occurs:

> **mapbox-gl.js:31 Error: The layer 'gl-draw-polygon-fill-inactive.cold' does not exist in the map's style and cannot be queried for features.**
   > at i.queryRenderedFeatures (mapbox-gl.js:35:133328)
   > at r.queryRenderedFeatures (mapbox-gl.js:35:323577)
   > at ot (mapbox-gl-draw.js:1:4660)
   > at Object.click (mapbox-gl-draw.js:1:4319)
   > at rt (mapbox-gl-draw.js:1:5068)
   > at i.mousemove (mapbox-gl-draw.js:1:22101)

mapbox-gl-js version 1.13.2
mapbox-gl-draw version 1.3.0, 1.4.2

A suggested solution might be to update the featuresAt function to filter the queryParams.layers to only use layers that exist on the map:

`if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id);`

to

`if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id).filter(id => ctx.map.style.getLayer(id) != null);`

caldwellc:
Fixes #1183

Updated fetures_at.js's `featuresAt` function to verify a style layer exists before adding it to queryParam.layers for the map.queryRenderedFeatures function. The mapbox-gl-draw style layers are removed during a map.setStyle call and are not added back until the data event (with type of style) is received. If the featuresAt function includes missing styles, the map.queryRenderedFeatures function will fire an error (https://github.com/mapbox/mapbox-gl-js/blob/11ea4f82e2d04041ef33339ac51909e876eed910/src/style/style.js#L1111C1-L1112C1)

Updated test/utils/create_map.js mock of mapbox-gl-js's map to include a very basic style layer management.

Updated features_at.test.js to better mock the source/style handling of mapbox-gl-js so that a test case could be added to demonstrate this fix.

### DIFF
--- a/src/lib/features_at.js
+++ b/src/lib/features_at.js
@@ -29,7 +29,8 @@ function featuresAt(event, bbox, ctx, buffer) {
   const box = (event) ? mapEventToBoundingBox(event, buffer) : bbox;
 
   const queryParams = {};
-  if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id);
+
+  if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id).filter(id => ctx.map.style.getLayer(id) != null);
 
   const features = ctx.map.queryRenderedFeatures(box, queryParams)
     .filter(feature => META_TYPES.indexOf(feature.properties.meta) !== -1);
--- a/src/lib/features_at.js
+++ b/src/lib/features_at.js
@@ -30,7 +30,7 @@ function featuresAt(event, bbox, ctx, buffer) {
 
   const queryParams = {};
 
-  if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id).filter(id => ctx.map.style.getLayer(id) != null);
+  if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id).filter(id => ctx.map.getLayer(id) != null);
 
   const features = ctx.map.queryRenderedFeatures(box, queryParams)
     .filter(feature => META_TYPES.indexOf(feature.properties.meta) !== -1);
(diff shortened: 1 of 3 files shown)


src/lib/features_at.js
```diff
@@ -29,7 +29,8 @@ function featuresAt(event, bbox, ctx, buffer) {
   const box = (event) ? mapEventToBoundingBox(event, buffer) : bbox;
 
   const queryParams = {};
-  if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id);
+
+  if (ctx.options.styles) queryParams.layers = ctx.options.styles.map(s => s.id).filter(id => ctx.map.style.getLayer(id) != null);
```
stepankuzmin, line 33: Nit: preferably use `map.getLayer` for getting layer

stepankuzmin:
Hi @caldwellc,

Thanks for the contribution! Overall this looks good to me, just one minor nit.

src/lib/features_at.js
caldwellc: Sounds good, will make that change.

lemmingapex:
Any ideas when we can expect a 1.4.3 release @stepankuzmin?  Thanks!

stepankuzmin:
Hi @lemmingapex. I just published v1.4.3 https://github.com/mapbox/mapbox-gl-draw/releases/tag/v1.4.3

PR #1210, opened 2024-02-20
Migrate to GitHub Actions

stepankuzmin:
Migrating from TravisCI to GitHub Actions for testing. Also, add CODEOWNERS.


.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
```
underoot, line 14: [nit] Not sure that we need node.js version matrix for browser tests if we don't have plan to launch it actually for several versions

.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'yarn'
+
+    - run: yarn --frozen-lockfile
```
underoot, line 26: Shall we add cache for dependencies?

underoot (changes requested):

.github/workflows/main.yml
stepankuzmin: Done

underoot (approved):

PR #1191, opened 2023-08-03
Draw should not interfere with dragPan.disable()

MauPalantir:
Fixes #654

The issue is that Draw assumes that dragPan is enabled and simply enables it without question after certain operations (when dragpan is temporarily disabled). This change would set a variable called initialDragPanState and only re-enables when it was originally enabled.

PR #1216, opened 2024-03-14
Draw should not interfere with dragPan.disable() 

itoxiq:
Continuation of #1191

The issue is that Draw assumes that dragPan is enabled and simply enables it without question after certain operations (when dragpan is temporarily disabled). This change would set a variable called initialDragPanState and only re-enables when it was originally enabled.


iToXiQ:
@stepankuzmin

stepankuzmin (approved):
Looks good to me, thanks for the contribution @itoxiq 👍

fede-qubika:
@itoxiq @stepankuzmin hey guys, after clicking the same feature multiple times without moving the mouse, doesn't it happen to you that dragPan remains disabled, regardless of whether it was initially enabled? I just updated to version 1.5.0, and this issue started happening

underoot:
@fede-qubika The issue was addressed. Thank you for the report!

Issue #1143, opened 2023-01-13
Option to increase geojson maxzoom?


omasback:
I'm building a tool to draw features with millimeter precision. Unfortunately the default maxzoom of geoJson sources is 18, which equates to a rendering precision of about 10mm. 

I believe the offending line is here:
https://github.com/mapbox/mapbox-gl-draw/blob/c12fb64b90d00e877e94b12b60bc3b80e42924dd/src/setup.js#L83

Im pretty sure this could be fixed by setting maxzoom of geojson data source to >=24, or even making it an init option.

Issue #858, opened 2019-01-12
Unable to use line-gradient paints styles for lines


karlTGA:
**mapbox-gl-js version**: 0.52
**mapbox-gl-draw version**: 1.1.1

### Steps to Trigger Behavior

 1. Set a line-gradient value in a paint object of a line style.
``` javascript 
{
    id     : 'gl-draw-line-static-gradient',
    type   : 'line',
    filter : ['all', ['==', '$type', 'LineString'], ['==', 'mode', 'static']],
    paint  : {
      'line-color'   : 'red',
      'line-width'   : 10,
      'line-gradient': ['interpolate', ['linear'], ['line-progress'], 0, 'blue', 0.1, 'royalblue', 0.3, 'cyan', 0.5, 'lime', 0.7, 'yellow', 1, 'red'],
    },
    layout: {
      'line-cap': 'round',
      'line-join': 'round',
    },
  },
```
### Expected Behavior

Get a nice line with a gradient from green to red. 

### Actual Behavior

Get an error message:
Error: layers.gl-draw-line-static-gradient.cold: layer "gl-draw-line-static-gradient.cold" specifies a line-gradient, which requires a GeoJSON source with `lineMetrics` enabled.
    at Object.Jr [as emitValidationErrors] (mapbox-gl.js:29)
    at De (mapbox-gl.js:33)
    at i._validate (mapbox-gl.js:33)
    at i.addLayer (mapbox-gl.js:33)
    at r.addLayer (mapbox-gl.js:33)
    at ctx.options.styles.forEach.style (setup.js:102)
    at Array.forEach (<anonymous>)
    at Object.addLayers (setup.js:101)
    at r.connect (setup.js:35)
    at r.ht.fire (mapbox-gl.js:29)

andrewharvey:
When the GeoJSON source is added to the map at https://github.com/mapbox/mapbox-gl-draw/blob/5b2280817c901d93b05c937a7bdb2efb8afefe0c/src/setup.js#L84-L99 you need to use 'lineMetrics: true'.

Mapbox GL Draw could probably provide an option for this so you can easily enable it.

karlTGA:
Aha ok.
Maybe you can add an optional parameter to the `add` function and give the options to mapbox through.  
`add(geojson: Object, options?: Object)`
So every special option is useable.

gtsl-2:
I think lineMetrics works with the relatively new mapbox-gl version.
[...]

PR #1218, opened 2024-04-21
Add hotSourceOptions and coldSourceOptions

shushanyalida5124:
Add hotSourceOptions and coldSourceOptions to the options to pass to the drawing and drawn layers. This could fix issue [#858](https://github.com/mapbox/mapbox-gl-draw/issues/858) and [#1143](https://github.com/mapbox/mapbox-gl-draw/issues/1143).


swevens (approved):

lukasvdelft:
if this would be implemented i could switch form leaflet to mapbox as I need the precision (high zoomlevels).

PR #1242, opened 2024-06-27
Modernize 2024

stepankuzmin:
* Switch to ESM
* Switch to NPM instead of Yarn
* Switch to native Node test runner
* Drop the `buble` compiler ~~and switch to the `esbuild`~~
* Drop `xtend`, use `Object.assign` instead
* Drop `lodash.isequal`, use `fast-deep-equal` instead
* Use `vite` for the local development
* Update the debug/ page: use GL JS and GL Geocoder from NPM so we can keep them up to date with Dependabot

The minified bundle size dropped from 80K to 68K.

The benchmarks primarily use the previous setup.


mourner (commented):
Nice! Why do we need ESBuild though if we still use Rollup?

stepankuzmin:
You are right, we don't need to transpile with esbuild for older browsers anymore 👍

bench/README.md
```diff
@@ -7,13 +7,12 @@ Benchmarks help us catch performance regressions and improve performance.
 Start the benchmark server
 
 ```bash
-MAPBOX_ACCESS_TOKEN={YOUR MAPBOX ACCESS TOKEN} npm start
+MAPBOX_ACCESS_TOKEN={YOUR MAPBOX ACCESS TOKEN} yarn start-bench
```
mourner, line 10: Let's drop Yarn in favor of NPM too

mourner (approved):
Nice! Something I've been doing recently is also removing tape and nyc in favor of native Node test runner and coverage — see if it makes sense to do so here, but not too important.

underoot (approved):

PR #1248, opened 2024-07-02
chore: Update package.json to Node.js version 18.0.0 or higher

varna:
https://github.com/mapbox/mapbox-gl-draw/pull/1242 switched project to Vite and ESM. This means that this project currently supports node version `^18.0.0 || >=20.0.0` https://vitejs.dev/guide/#scaffolding-your-first-vite-project

By adding `engines.node` to `package.json` we gain some extra features:

- Can automatically create a warning, when someone uses a wrong version
- Automatically use defined Node version, when running project via pnpm
- It will probably also be picked up by CI/CD

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "@mapbox/mapbox-gl-draw",
   "version": "1.4.3",
+  "engines": {
+    "node": "^18.0.0 || >=20.0.0"
+  },
   "description": "A drawing component for Mapbox GL JS",
   "author": "mapbox",
   "license": "ISC",


stepankuzmin (approved):
Good catch, thanks @varna 👍

mourner:
Note that this requirement is only for local development of `mapbox-gl-draw` — downstream consumers that bundle `mapbox-gl-draw` in their apps don't have to adhere to the requirement. This is why I've been avoiding specifying `engines` for such projects. Not sure if we want to keep it.

varna:
Oh well, then shall we go with `.nvmrc` instead?  @mourner 

I thought `engines` was not enforced without `engines-strict`. And `vite` has `engines` set too. So downstream user would need to use Draw as dev dependency anyway then (and bundler). Or via CDN. In theory Draw should work with node14 (es2020). But node18 is actually the latest maintained version (and the one used by Vite), so I wouldn't recommend using anything older.

image

varna:
Actually, this project doesn't even really need Node to run. Currently only dependence is probably node test runner. It should be easy to replace it with Vitest. Then you could use this as ES on anything (Node, Bun or Deno).

PR #1274, opened 2024-08-15
Use native `crypto.randomUUID()` instead of `hat` module

varna:
Reducing number of dependencies by replacing `hat` module with native `crypto.randomUUID()` which is well supported:

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility
https://caniuse.com/mdn-api_crypto_randomuuid

This would push `mapbox-gl-draw` dev env to node 20 (or deno/bun). Node 18 is reaching end of maintenance next year, so idk if that's an issue for you.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,8 +12,7 @@
         "@mapbox/geojson-area": "^0.2.2",
         "@mapbox/geojson-normalize": "^0.0.1",
         "@mapbox/point-geometry": "^0.1.0",
-        "fast-deep-equal": "^3.1.3",
-        "hat": "0.0.3"
+        "fast-deep-equal": "^3.1.3"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


underoot:
Hello, @varna! Thank you for your contribution! To me, it seems that replacing of hat with crypto.randomUUID() will cause inability to use maps [on site without HTTPS](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility). I agree that `hat` module seems outdated and need to be replaced. Maybe we can consider some another alternatives like [`nanoid`](https://github.com/ai/nanoid?tab=readme-ov-file#comparison-with-uuid)

cc @stepankuzmin

underoot:
Thank you for contribution! We decided to go with nanoid because of non-secure contexts #1318

PR #1275, opened 2024-08-15
fix: update exports to work with current release workflow

varna:
At the moment, the release version of this package doesn't work anymore when imported into another project.
This fix allows ESM imports to work again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,12 @@
     "drawing"
   ],
   "type": "module",
-  "exports": "./index.js",
+  "exports": {
+    ".": "./index.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -22,7 +22,12 @@
     "drawing"
   ],
   "type": "module",
-  "exports": "./index.js",
+  "exports": {
+    ".": "./index.js",
```
underoot, line 26: I think it should point to `./dist/mapbox-gl-draw.js`

stepankuzmin (approved):

PR #1394, opened 2025-04-26
refactor: use `key` with fallback to `keyCode` for compatibility

erotourtes:
The `keyCode` property [is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and may, or may not be supported in modern browsers

This PR updates functions to check for `key` property and fallback to `keyCode` for compatibility with older browsers that don't support the key property.

### DIFF
--- a/src/events.js
+++ b/src/events.js
@@ -1,9 +1,10 @@
-import setupModeHandler from './lib/mode_handler.js';
-import getFeaturesAndSetCursor from './lib/get_features_and_set_cursor.js';
+import * as Constants from './constants.js';
 import featuresAt from './lib/features_at.js';
+import getFeaturesAndSetCursor from './lib/get_features_and_set_cursor.js';
+import { CommonSelectors } from './lib/index.js';
 import isClick from './lib/is_click.js';
 import isTap from './lib/is_tap.js';
-import * as Constants from './constants.js';
+import setupModeHandler from './lib/mode_handler.js';
 import objectToMode from './modes/object_to_mode.js';
 
 export default function(ctx) {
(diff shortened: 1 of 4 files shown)


src/events.js
```diff
@@ -1,9 +1,10 @@
-import setupModeHandler from './lib/mode_handler.js';
-import getFeaturesAndSetCursor from './lib/get_features_and_set_cursor.js';
+import * as Constants from './constants.js';
 import featuresAt from './lib/features_at.js';
+import getFeaturesAndSetCursor from './lib/get_features_and_set_cursor.js';
+import { CommonSelectors } from './lib/index.js';
```
underoot, line 4: I'd prefer direct import to prevent potential circular imports

src/events.js
```diff
@@ -120,30 +121,36 @@ export default function(ctx) {
...
   };
 
-  // 8 - Backspace
-  // 46 - Delete
-  const isKeyModeValid = code => !(code === 8 || code === 46 || (code >= 48 && code <= 57));
+  const isKeyModeValid = (event) => {
+    const isBackspaceKey = CommonSelectors.isBackspaceKey(event);
+    const isDeleteKey = CommonSelectors.isDeleteKey(event);
+
+    const key = event.key || String.fromCharCode(event.keyCode);
```
underoot, line 128: Candidate to extract to `CommonSelectors` as well, wdyt?

underoot (commented):
Thank you for your contribution! Left some comments, overall looks good!

src/events.js
erotourtes: Done - good call

erotourtes (commented):
Thank you for taking the time to review this!

underoot (approved):

PR #1428, opened 2025-07-28
Fix: In direct_select mode it's better to prefer selecting vertices than midpoints

denull:
When a polygon or a polyline contains two vertices very close to each other, it can be very tricky to click on one of them and not on the midpoint between them.

The result is very frustrating, because this would create yet another vertex (and if user decides to click multiple times, it would produce even more vertices). This issue is even worse on touch devices (because `touchBuffer` is much bigger than `clickBuffer`).

This commit fixes the issue by tweaking the sort order of points: if there are more than one vertex near the cursor, we should always try to select an existing vertex instead of creating a new one.

### DIFF
--- a/src/lib/sort_features.js
+++ b/src/lib/sort_features.js
@@ -8,11 +8,22 @@ const FEATURE_SORT_RANKS = {
   Polygon: 2
 };
 
+const POINT_META_RANKS = {
+  vertex: 0,
+  midpoint: 1,
+  feature: 2
+};
+
 function comparator(a, b) {
   const score = FEATURE_SORT_RANKS[a.geometry.type] - FEATURE_SORT_RANKS[b.geometry.type];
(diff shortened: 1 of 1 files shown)


underoot:
Hey, @denull. Thank you for your contribution! It's a nice improvement! Could you share the video before and after to visually demonstrate the difference in the behaviour? 

I'm also considering integration tests, but unfortunately, we haven't set up anything here yet: only some tests that we run in a Node.js environment with a mocked browser. But I guess we can do it later as a follow-up

denull:
Btw, yet another potential improvement would involve sorting vertices based on their distance from the actual cursor position — similar to how polygons are sorted based on their area now. But that would probably require adding an extra dependency (cheap-ruler, for example), so I decided not to include it in this PR.

PR #1491, opened 2025-12-08
Reset drag-pan state on stoping DirectSelect mode

leonkenneth:
Right now if the mode is changed while a drag is still ongoing, the dragpan is left disabled.

We have the case in our application. That may be be a peculiarity of our code ; but I thought it makes sense to reset this state on exiting the mode.


underoot (approved):
LGTM!

PR #1501, opened 2026-01-15
Avoid duplicate layers when a style already defines a source

0xOnyx:
This PR fixes an issue in options.js where styles with a predefined source were added twice, producing duplicate layers with the same id. That breaks Mapbox GL because layer IDs must be unique.
Now, styles that already define a source are kept once, and only styles without a source are duplicated for cold and hot.
Why
When a style already has a source, it should not be duplicated. The previous behavior created duplicate layer IDs and caused errors for users who define their own source.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -37,13 +37,10 @@ const hideControls = {
 };
 
 function addSources(styles, sourceBucket) {
-  return styles.map((style) => {
-    if (style.source) return style;
-    return Object.assign({}, style, {
-      id: `${style.id}.${sourceBucket}`,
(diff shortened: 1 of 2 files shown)


0xOnyx:
Is it possible to approve/accept the PR?

underoot (dismissed):
LGTM, sorry for being slow with the review.
Will add tests and merge. Thank you for your patience!

underoot (approved):

stepankuzmin (approved):
Thanks!

